PR #1, opened 2015-03-27
(TK-170) Inital setup of trapperkeeper status service project

rlinehan:
Add skeleton derived from `lein new default puppetlabs/trapperkeeper-status` (so that we have a test to run) and enable travis.


KevinCorcoran:
LOOKS SO GOOD

:eyes:

PR #2, opened 2015-04-16
(TK-171) Initial implementation of status service

rlinehan:
Add a `register-status` protocol and function used by other services to register a callback function with the status service.

Add a `/services` rollup status endpoint that returns the status of all services that have registered a callback function.


cprice404:
+1, looking good.  I left a few nitpick questions / comments, but nothing that I'd consider a blocker for merge.

rlinehan:
@cprice404 Added a new commit addressing two of your comments. As far as the `make-fn-schema`, would it be okay to leave adding the proper representation for inputs/outputs until we do a little more with it (I'm thinking when we add support for querying based on status level)? I could also add a TODO back in, if you'd like.

cprice404:
+1.  I'm fine merging the make-fn-schema w/o having finalized the actual schema there, your plan sounds fine.  My preference would be to include a TODO just to sort of help make sure it doesn't slip through the cracks later, but that's not critical.  If someone else reviews this and +1's it before a TODO comment gets added, or if you prefer not to add one, we should just go ahead and merge this w/o it.

rlinehan:
Added a commit adding a TODO.

src/puppetlabs/trapperkeeper/services/status/status_service.clj
```diff
@@ -0,0 +1,30 @@
...
+(defservice status-service
+  StatusService
+  [[:WebroutingService add-ring-handler]]
+
+  (init [this context]
+    (assoc context :status-fns (atom {})))
+
+  (start [this context]
+    (log/info "Registering status service HTTP API at /status")
+    (let [handler (core/build-handler context)]
```
KevinCorcoran: I'd prefer the status functions to be extracted from `context` here and then passed-in to `build-handler`, as opposed to passing the TK context around.  It looks a bit confusing (especially since compojure has its own context) and the core functions just need those functions, so seeing that in arg lists makes it a bit more obvious what's going on.  Yes, this is a nitpick.  :nail_care:

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -0,0 +1,73 @@
...
+  (let [latest-status (last (sort-by :service-status-version service))]
+    {:service-version (:service-version latest-status)
+     :service-status-version (:service-status-version latest-status)
+     :status ((:status-fn latest-status))}))
+
+(schema/defn call-status-fns :- ServicesStatus
+  [context]
+  "Call the latest status function for each service in the service context,
+  and return a map of service to service status."
+  (ks/mapvals call-latest-status-fn-for-service (deref (:status-fns context))))
```
KevinCorcoran: I wonder if it's worth using `pmap` here.

nwolfe: But you're locked in to doing it the kitchensink way! :smirk: 

Interestingly, I was just looking at the analogous `get-certificate-statuses` endpoint in the CA wondering what it would be like if we used `pmap`.

src/puppetlabs/trapperkeeper/services/status/status_service.clj
```diff
@@ -0,0 +1,30 @@
+(ns puppetlabs.trapperkeeper.services.status.status-service
+  (:require [clojure.tools.logging :as log]
+            [compojure.core :as compojure]
+            [puppetlabs.trapperkeeper.core :refer [defservice]]
+            [puppetlabs.trapperkeeper.services :refer [service-context]]
+            [puppetlabs.trapperkeeper.services.status.status-core :as core]
+            [puppetlabs.trapperkeeper.services.webrouting.webrouting-service :as webrouting-service]))
+
+(defprotocol StatusService
+  (register-status [this service-name service-version status-version status-fn]
```
KevinCorcoran, line 9: I know that this corresponds to the design doc, but I'd like `service-version` and `status-version` to be optional.  Thoughts?

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -0,0 +1,73 @@
...
+
+(schema/defn call-latest-status-fn-for-service :- ServiceStatus
+  "Given a list of maps containing service information and a status function,
+  find the latest version, and return a map with the service's version, the
+  version of the service's status, and the results of calling this status
+  function."
+  [service :- [ServiceInfo]]
+  (let [latest-status (last (sort-by :service-status-version service))]
+    {:service-version (:service-version latest-status)
+     :service-status-version (:service-status-version latest-status)
```
KevinCorcoran: These two lines can be written a bit more concisely using `select-keys`.

src/puppetlabs/trapperkeeper/services/status/status_service.clj
KevinCorcoran, line 9: Also, I think it should be `register-status!` since it's just called for side-effects.

src/puppetlabs/trapperkeeper/services/status/status_service.clj
```diff
@@ -0,0 +1,30 @@
+(ns puppetlabs.trapperkeeper.services.status.status-service
+  (:require [clojure.tools.logging :as log]
+            [compojure.core :as compojure]
+            [puppetlabs.trapperkeeper.core :refer [defservice]]
+            [puppetlabs.trapperkeeper.services :refer [service-context]]
+            [puppetlabs.trapperkeeper.services.status.status-core :as core]
+            [puppetlabs.trapperkeeper.services.webrouting.webrouting-service :as webrouting-service]))
```
KevinCorcoran: This namespace does not need to be `require`d.

KevinCorcoran:
:chart_with_upwards_trend: looks good overall.

src/puppetlabs/trapperkeeper/services/status/status_service.clj
rlinehan, line 9: I agree that `service-version` should be optional, and that it might be useful to have helper functions that can pull out the service's version from the lein metadata.

I'm not so sure about `status-version`. Compared to `service-version`, actually supplying this as an argument seems pretty easy, and it's not going to change as often as `service-version`. If `status-version` is optional but someone registers two different status functions for one service, what happens (this is actually also potentially a current problem - two status functions can be registered without changing the status version... )? Do we just register both of them and choose randomly which to call for the roll-up endpoint? Or do we only register one of them?

KevinCorcoran, line 9: > it might be useful to have helper functions that can pull out the service's version from the lein metadata.

Yeah, I think we'll probably want to do something like this eventually.

re: making `status-version` optional ...

I think we could come up with a reasonable default behavior for the case when no version is provided but multiple functions are registered.  Perhaps we just throw an error which tells the caller that, if you want to do that, you need to specify versions.  Or, we could just say that "last one wins", although I think throwing an error is probably better.  Either way, this issue falls into my bucket of "reasonable defaults" which is one of my favorite programming topics to rant about in code review.  :bicyclist: :smirk_cat: 

> this is actually also potentially a current problem - two status functions can be registered without changing the status version... 

I think this case should probably throw an error as well.  The only thing I can imagine that would cause this to happen would be a bug - I can't come up with any good reason for wanting to overwrite a previously-registered status function.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
rlinehan: Yes, I think it is worth using `pmap` - I messed about some, and I think it will work well here assuming some status-fns will take a little more time. Sidenote: I didn't really understand `pmap` before, but thanks to the Clojure/West talk on parallelism I do!

src/puppetlabs/trapperkeeper/services/status/status_service.clj
rlinehan, line 9: I think everything you've said makes sense - sensible defaults and throwing an error in both those cases. Do you think that should be done as part of this ticket? (I think it does? but I could be convinced otherwise... I just don't really want to make another JIRA ticket.)

src/puppetlabs/trapperkeeper/services/status/status_core.clj
KevinCorcoran: I could go either way on `map` or `pmap` here ... it might be an optimization for some cases but we obviously can't say for sure.  I lean towards `pmap` since it might be slower in the fast cases but has the potential to be much faster in the pathologically-slow cases.

src/puppetlabs/trapperkeeper/services/status/status_service.clj
KevinCorcoran, line 9: Let's shelve it for now.  I don't want to diverge from the design doc without @cprice404's sign-off.  I created https://tickets.puppetlabs.com/browse/TK-210.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
rlinehan: I did some benchmarking, and I think that as soon as the status-fn starts doing a little bit more work for one service and there is more than one service registered that `pmap` will be faster. It seems like we have quite a few cases where more than one service will be registered and where we'll potentially want to be returning a descent amount of information.

In the case where `pmap` is faster it's about twice as fast; in the case where it's not it's about 0.13 msecs vs. 0.07 msecs. So basically, I think that while `pmap` isn't always an optimization but when it is slower it's by very little, as you said in the pathologically-slow cases it will make a huge difference.

KevinCorcoran: Lies, damned lies, and benchmarks...

Nah, I kid, thanks for poking at this.  :smiling_imp:

rlinehan:
@KevinCorcoran I think I've addressed all your feedback.

KevinCorcoran:
:dart:

PR #3, opened 2015-04-24
(TK-172) Add endpoint to query status for a single service

rlinehan:
Add the ability to query for a single service's status at
/status/v1/services/<service name>. If the service has multiple status
callback functions registered, the endpoint returns the result of calling the
latest (same as for the rollup endpoint). If the service is nonexistent/hasn't
registered itself with the status service, the endpoint returns a 404.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/status/status_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_core.clj
@@ -70,12 +70,12 @@
           {:status 200
            :body statuses}))
        (compojure/GET "/services/:service-name" [service-name]
-         (if-let [status-info (get (deref status-fns-atom) service-name)]
-           (let [status (call-latest-status-fn-for-service status-info)]
+         (if-let [service-info (get (deref status-fns-atom) service-name)]
+           (let [status (call-latest-status-fn-for-service service-info)]
              {:status 200
               :body (assoc status :service-name service-name)})
            {:status 404
-            :body {:error {:type ::service-not-found
+            :body {:error {:type :service-not-found
                            :message (str "No status information found for service "
                                          service-name)}}})))))
 
--- a/src/puppetlabs/trapperkeeper/services/status/status_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_core.clj
@@ -68,7 +68,15 @@
       (compojure/GET "/services" []
         (let [statuses (call-status-fns status-fns-atom)]
           {:status 200
-           :body {"services" statuses}})))))
+           :body {"services" statuses}}))
+       (compojure/GET "/services/:service-name" [service-name]
+         (if-let [status-info (get (deref status-fns-atom) service-name)]
+           {:status 200
+            :body (call-latest-status-fn-for-service status-info)}
+           {:status 404
+            :body {:error {:type ::service-not-found
+                           :message (str "No status information found for service "
+                                         service-name)}}})))))
 
 (defn build-handler [status-fns-atom]
   (-> (build-routes status-fns-atom)
--- a/src/puppetlabs/trapperkeeper/services/status/status_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_core.clj
@@ -71,8 +71,9 @@
            :body {"services" statuses}}))
        (compojure/GET "/services/:service-name" [service-name]
          (if-let [status-info (get (deref status-fns-atom) service-name)]
-           {:status 200
-            :body (call-latest-status-fn-for-service status-info)}
+           (let [status (call-latest-status-fn-for-service status-info)]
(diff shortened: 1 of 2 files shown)


test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -45,3 +45,27 @@
...
+     foo-service]
+    {:webserver {:port 8180
+                 :host "0.0.0.0"}
+     :web-router-service {:puppetlabs.trapperkeeper.services.status.status-service/status-service "/status"}}
+    (testing "returns service information for service that has registered a callback"
+      (let [req (http-client/get "http://localhost:8180/status/v1/services/foo")]
+        (is (= 200 (:status req)))
+        (is (= {"service-version" "1.1.0"
+                "service-status-version" 2
+                "status" "foo status 2"}
```
rlinehan: The only thing I wasn't sure about here was whether it would be useful to also return the service's name in the body?

MSLilah:
:+1:

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
jeffmccune: Name might be useful to allow consumers to interact only with the data returned, not their own context or state.

MSLilah: I'd lean towards including it. I think the user can definitely get by without it since they'll need it anyway to query a specific service's status, but I think it might be a little more convenient to have it in the body.

rlinehan:
Added a commit adding the service name to the response body for the single service endpoint.

Also added a commit removing the outer map which only contained one key (`services`) from the rollup endpoint.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -68,7 +68,16 @@
       (compojure/GET "/services" []
         (let [statuses (call-status-fns status-fns-atom)]
           {:status 200
-           :body {"services" statuses}})))))
+           :body statuses}))
+       (compojure/GET "/services/:service-name" [service-name]
+         (if-let [status-info (get (deref status-fns-atom) service-name)]
```
KevinCorcoran: I think a better name would be `service-info`, the corresponding schema is `ServiceInfo`.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -68,7 +68,16 @@
...
+           (let [status (call-latest-status-fn-for-service status-info)]
+             {:status 200
+              :body (assoc status :service-name service-name)})
+           {:status 404
+            :body {:error {:type ::service-not-found
```
KevinCorcoran: I'm not sure if I like namespaces ending up in the response.  Perhaps just `:service-not-found` here instead?

KevinCorcoran:
:+1: Looks fine overall.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
rlinehan: I agree, this is just what we've done in other places (e.g. file sync)?

KevinCorcoran: ORLY?  Well, I'd be happy to change it in file sync, and I wouldn't consider anything in that repo to be canon at this point.

rlinehan: https://github.com/rlinehan/pe-file-sync/commit/fff3b910900f14452a6faba23042f73e3b859849#commitcomment-10500725 and https://github.com/rlinehan/pe-file-sync/commit/dfc313bf473b57e1dd3000d5c5ebf44ed0bb70e9, although in that case I was kind of making up a partial namespace. I think sometimes having the response indicate what trapperkeeper service an error came from could maybe be useful in the case of http responses (e.g. `:status-service/service-not-found`), but I don't like having the full namespace in the response - I'd prefer just `:service-not-found` rather than the full namespace.

KevinCorcoran: Hmm, yeah, I can see something like that being useful in general, but ...

In this specific case, a user doesn't care about the status service at all - hopefully, they don't even know it exists.  They'll be querying for the status of an individual service they care about, e.g. Puppet Server, so anything about `status-service` in the response is probably just noise (at least in this use-case).

I feel like we're treading on :bike: territory here ... I vote for just `:service-not-found` but since we don't have a hard requirement one way or the other at this point, I'll be fine with whatever you want to go with.

rlinehan:
@KevinCorcoran added a commit addressing your comments.

PR #4, opened 2015-04-28
(TK-174) Query based on status level

rlinehan:
Add support for a `level` query param. When this query param is used, the
status function for the services are called with that level. If no level is
specified, defaults to `info`.

Also adds ringutils to help handle HTTP errors.

~~This is based off of #3 and will likely need to be rebased after that is in.~~ This PR has been rebased and should now be able to be merged.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/status/ringutils.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/ringutils.clj
@@ -1,0 +1,65 @@
+;; TODO: Much of this code is copied from other projects. It should
+;; probably be moved into a shared library or some such.
+(ns puppetlabs.trapperkeeper.services.status.ringutils
+  (:require [clojure.tools.logging :as log]
+            [slingshot.slingshot :refer [try+]]
+            [puppetlabs.kitchensink.core :as ks]))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Public
+
+(defn wrap-request-logging
+  "A ring middleware that logs the request."
+  [handler]
+  (fn [{:keys [request-method uri] :as req}]
+    (log/debug "Processing" request-method uri)
+    (log/trace "---------------------------------------------------")
+    (log/trace (ks/pprint-to-string (dissoc req :ssl-client-cert)))
+    (log/trace "---------------------------------------------------")
+    (handler req)))
+
+(defn wrap-response-logging
+  "A ring middleware that logs the response."
+  [handler]
+  (fn [req]
+    (let [resp (handler req)]
+      (log/trace "Computed response:" resp)
+      resp)))
+
+(defn wrap-request-data-errors
+  "A rignt middleware that catches slingshot errors of :type
+  :request-dava-invalid and returns a 400."
+  [handler]
+  (fn [request]
+    (try+ (handler request)
+          (catch [:type :request-data-invalid] e
+            {:status 400
+             :body {:error e}}))))
+
+(defn wrap-schema-errors
+  "A ring middleware that catches schema errors and returns a 500
+  response with the details"
+  [handler]
+  (fn [request]
+    (try (handler request)
+         (catch clojure.lang.ExceptionInfo e
+           (let [message (.getMessage e)]
+             (if (re-find #"does not match schema" message)
+               {:status 500
+                :body {:error {:type :application-error
+                               :message (str "Something unexpected happened: "
+                                             (:error (.getData e)))}}}
+               ;; re-throw exceptions that aren't schema errors
+               (throw e)))))))
+
+(defn wrap-errors
+  "A ring middleware that catches all otherwise uncaught errors and
+  returns a 500 response with the error message"
+  [handler]
+  (fn [request]
+    (try (handler request)
+         (catch Exception e
+           (log/error e "Error on server")
+           {:status 500
+            :body {:error {:type :application-error
+                           :message (str "Error on server: " e)}}}))))
--- a/src/puppetlabs/trapperkeeper/services/status/ringutils.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/ringutils.clj
@@ -9,7 +9,7 @@
 ;;; Public
 
 (defn wrap-request-data-errors
-  "A rignt middleware that catches slingshot errors of :type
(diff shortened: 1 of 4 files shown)


src/puppetlabs/trapperkeeper/services/status/ringutils.clj
```diff
@@ -0,0 +1,65 @@
+;; TODO: Much of this code is copied from other projects. It should
+;; probably be moved into a shared library or some such.
```
rlinehan, line 2: I copied this straight out of file sync service (which copied a lot of this out of puppet server). Perhaps we should add these to https://github.com/puppetlabs/ring-middleware?

KevinCorcoran, line 2: Ha, yeah ... mess proliferation.  Although it probably needs some cleanup before landing in ring-middleware.

rlinehan:
Updated/rebased and should now be ready to be reviewed and merged.

rlinehan:
Added commits to use `handler/api` and to remove unused middleware.

src/puppetlabs/trapperkeeper/services/status/ringutils.clj
```diff
@@ -0,0 +1,47 @@
...
+(ns puppetlabs.trapperkeeper.services.status.ringutils
+  (:require [clojure.tools.logging :as log]
+            [slingshot.slingshot :refer [try+]]
+            [puppetlabs.kitchensink.core :as ks]))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Public
+
+(defn wrap-request-data-errors
+  "A rignt middleware that catches slingshot errors of :type
```
KevinCorcoran: s/rignt/ring

KevinCorcoran:
:shipit:

src/puppetlabs/trapperkeeper/services/status/ringutils.clj
cprice404, line 2: Can we create a jira for this cleanup work?

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -63,22 +82,28 @@
...
-           {:status 404
-            :body {:error {:type :service-not-found
-                           :message (str "No status information found for service "
-                                         service-name)}}})))))
+  (handler/api
+    (compojure/routes
+      (compojure/context "/v1" []
+        (compojure/GET "/services" [:as {params :params}]
+          (let [level (get-status-detail-level params)
+                statuses (call-status-fns (deref status-fns-atom) level)]
```
cprice404, line 90: This is a little bit of a nitpick, and this wasn't introduced in this PR... but...

Looking over this namespace, it seems to me like the fact that there is an atom involved in our management of the state of the service is an implementation detail that doesn't necessarily need to bleed through here?

I _think_ it should always be the case that the registration of the status callback functions for various services will happen during lifecycle init phases, and thus the atom will be effectively immutable by the time we want to build the routes... in which case it might be better to just pass a plain-old map in to the route building fns?  And to the `update` function at the top of this namespace?  Then the use of the atom would be restricted to the service code, and since it's kind of tightly coupled with the TK service context, that makes some sense to me and seems like it might be a bit cleaner.

If you decide that's worth doing, feel free to merge after that change goes in, you don't need to wait another day for me to look at it again.

If you decide it's not worth doing, I'm good with this being merged w/o that change.

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -11,57 +11,83 @@
...
-      (is (= 200 (:status req)))
-      (is (= {"bar" {"service-version" "0.1.0"
-                     "service-status-version" 1
-                     "status" "bar status 1"}
-              "foo" {"service-version" "1.1.0"
-                     "service-status-version" 2
-                     "status" "foo status 2"}}
-             body)))))
+    (testing "returns latest status for all services"
+      (let [req (http-client/get "http://localhost:8180/status/v1/services")
```
cprice404: Maybe rename this var to `resp`?  I was confused momentarily because I thought you were reading the body of the request rather than the response.

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -11,57 +11,83 @@
...
+        (is (= 200 (:status req)))
+        (is (= {"bar" {"service-version" "0.1.0"
+                       "service-status-version" 1
+                       "status" "bar status 1 :info"}
+                "foo" {"service-version" "1.1.0"
+                       "service-status-version" 2
+                       "status" "foo status 2 :info"}}
+               body))))
+    (testing "uses status level from query param"
+      (let [req (http-client/get "http://localhost:8180/status/v1/services?level=debug")
```
cprice404: Same here.

cprice404:
+1 aside from minor comments.

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
rlinehan: Good point, that would make a lot more sense. I'll change that throughout this namespace.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
rlinehan, line 90: This makes a lot of sense, and I think it's worth doing. I'm not sure that this can be done for the `update` function (which gets called by `register-status!`) since it uses `swap!` to update the atom, but it could be unless changed to create a partial that gets used with `swap!` in the service code, or it could just be moved into the service code completely. In either of those changes the tests would probably need to change, but that's not a huge deal.

In any case, I think I'll open up a separate PR for these changes.

rlinehan:
I've added a PR changing `req` to `resp` per @cprice404's comments. I think I'll open a separate PR to  remove references to the atom from the core code.

KevinCorcoran:
@cprice404 I'm going to merge this so we can start some other work on top of it - it looks like all of your feedback has been addressed in one way or another.

src/puppetlabs/trapperkeeper/services/status/ringutils.clj
rlinehan, line 2: @cprice404 TK-214

PR #5, opened 2015-04-30
(TK-173) Support querying based on status-version

MSLilah:
Add support for querying based on a specific version of a
service's registered status function using the
service-status-version parameter.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/status/ringutils.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/ringutils.clj
@@ -14,7 +14,8 @@
   [handler]
   (fn [request]
     (try+ (handler request)
-          (catch [:type :request-data-invalid] e
+          (catch #(or (= :request-data-invalid (:type %))
+                      (= :service-status-version-not-found (:type %))) e
             {:status 400
              :body {:error e}}))))
 
--- a/src/puppetlabs/trapperkeeper/services/status/ringutils.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/ringutils.clj
@@ -14,8 +14,8 @@
   [handler]
   (fn [request]
     (try+ (handler request)
-          (catch #(or (= :request-data-invalid (:type %))
-                      (= :service-status-version-not-found (:type %))) e
+          (catch #(contains? #{:request-data-invalid :service-status-version-not-found}
+                             (:type %)) e
             {:status 400
              :body {:error e}}))))
 
--- a/src/puppetlabs/trapperkeeper/services/status/status_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_core.clj
@@ -98,12 +98,11 @@
    check whether it is valid. If not, throw an error."
   [params]
   (when-let [level (params :service-status-version)]
-    (try
-      (Integer. level)
-      (catch NumberFormatException e
-        (throw+ {:type :request-data-invalid
-                 :message (str "Invalid service-status-version. Should be an "
-                               "integer but was " level)})))))
+    (if-let [parsed-level (ks/parse-int level)]
+      parsed-level
+      (throw+ {:type    :request-data-invalid
+               :message (str "Invalid service-status-version. Should be an "
+                             "integer but was " level)}))))
 
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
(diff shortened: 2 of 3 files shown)


src/puppetlabs/trapperkeeper/services/status/ringutils.clj
```diff
@@ -14,7 +14,8 @@
   [handler]
   (fn [request]
     (try+ (handler request)
-          (catch [:type :request-data-invalid] e
+          (catch #(or (= :request-data-invalid (:type %))
+                      (= :service-status-version-not-found (:type %))) e
```
cprice404: This could be `#(contains? #{:request-data-invalid :service-status-version-not-found} (:type %))`, but, that's not a blocker

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -45,23 +45,40 @@
...
+    service-status-version :- (schema/maybe schema/Int)]
+    (let [status (if (nil? service-status-version)
+                   (last (sort-by :service-status-version service))
+                   (first (filter #(= (:service-status-version %)
+                                      service-status-version)
+                                  service)))]
+      (when (nil? status)
+        (throw+ {:type    :service-status-version-not-found
+                 :message (str "No status function with version "
+                               service-status-version
```
cprice404, line 70: this is going to log a `nil` if they didn't specify a status version and there are no callbacks registered for the service, correct?  Or do we already have a guard elsewhere that guarantees that we won't get here if they haven't registered a callback?

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -76,6 +93,18 @@
...
     :info))
 
+(defn get-service-status-version
+  "Given a params map from a request, get out the service status version and
+   check whether it is valid. If not, throw an error."
+  [params]
+  (when-let [level (params :service-status-version)]
+    (try
+      (Integer. level)
+      (catch NumberFormatException e
```
cprice404: I think we have a parse-int fn in kitchensink that does basically this, but if we don't already have a dep on KS it's probably not worth introducing one just for that.

cprice404:
LGTM, +1

src/puppetlabs/trapperkeeper/services/status/status_core.clj
rlinehan: Given that trapperkeeper itself has a dependency on KS, won't we always have a KS dependency? Or are you saying we don't want to have a dependency on a specific version of KS if we don't have to?

src/puppetlabs/trapperkeeper/services/status/status_core.clj
rlinehan, line 70: It will only be `nil` if a service version wasn't supplied. In that case, there must be at least one status function registered for the service, otherwise we'll return a `404` because we can't find anything for the service name. So I think this is fine.

rlinehan:
:+1: looks good! Seems like it would be slightly easier to make the two small changes that @cprice404 mentioned before merging, but we could also open up a maint PR after this gets merged - I'm fine with either.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
cprice404: oh, duh.  Yeah probably worth just using the fn from ks then.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
cprice404, line 70: k

MSLilah:
@cprice404 @rlinehan I've addressed your comments!

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -76,6 +93,18 @@
                 :message (str "Invalid level: " level)}))
     :info))
 
+(defn get-service-status-version
+  "Given a params map from a request, get out the service status version and
+   check whether it is valid. If not, throw an error."
+  [params]
+  (when-let [level (params :service-status-version)]
+    (let [parsed-level (ks/parse-int level)]
+      (if (nil? parsed-level)
```
rlinehan: I think this could be simplified by doing `(if-let [parsed-level (ks/parse-int level)]`

MSLilah:
@rlinehan I simplified those lines into an `if-let` statement

rlinehan:
dear travis, you are taking way too long.

PR #6, opened 2015-05-05
(TK-216) Fix hard-coded url prefix

cprice404:
The status service had a hard-coded path of `/status` that it
was passing to compojure's `context` macro, but was using
the web routing service to find the path where we mounted in
Jetty.  This meant that it was possible to configure the
web routing service to something other than `/status`, but
if you did so, the service would not be accessible.


rlinehan:
:+1: sorry I totally missed this before

PR #7, opened 2015-05-05
(maint) Pass status functions to routes, not atom

rlinehan:
The `status-fns` atom in the status service's context should be updated uring
that service's init. Thus, by the time start is called it should be immutable.
This commit moves the dereferencing of the atom into `start` in the service
namespace, so that the route-related functions in the core namespace only deal
with a map.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/status/status_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_core.clj
@@ -109,18 +109,17 @@
 ;;; Compojure App
 
 (defn build-routes
-  [status-fns-atom]
+  [status-fns]
   (handler/api
     (compojure/routes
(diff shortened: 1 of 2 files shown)


rlinehan:
@cprice404 this addresses the comments you left in https://github.com/puppetlabs/trapperkeeper-status/pull/4#discussion_r29412305. I think it's a bit more work to do the same thing for the `update-status-context` function in the core namespace, since that is doing a `swap!` (it would mean moving some tests around, at least), but I'm willing to do that here or to revisit that at some point if you think it's a good idea.

cprice404:
+1, I think this is an improvement regardless of what we decide to do about the `update` fn.

MSLilah:
:+1:

PR #8, opened 2015-05-06
(TK-210) Add utility function for getting service version

cprice404:
Because `service-version` is a required field for registering
a callback, and because we don't want to force all of the
consumers of this service to maintain a hard-coded instance
of that value by hand every time they do a new release of their
code, this commit introduces a utility function called
`get-service-version`.  If passed a valid maven group id and
artifact id, it will get the version string for that component.

This is intended to be used by downstream services; they should
call this prior to registering their callback, and then pass in
this value as the `service-version`.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/status/status_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_core.clj
@@ -5,7 +5,9 @@
             [ring.middleware.json :as ring-json]
             [slingshot.slingshot :refer [throw+]]
             [puppetlabs.kitchensink.core :as ks]
-            [puppetlabs.trapperkeeper.services.status.ringutils :as ringutils]))
+            [puppetlabs.trapperkeeper.services.status.ringutils :as ringutils]
+            [clj-semver.core :as semver]
+            [trptcolin.versioneer.core :as versioneer]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Schemas
@@ -30,8 +32,29 @@
 (def ServicesStatus
   {schema/Str ServiceStatus})
 
+(def SemVerVersion
+  (schema/pred semver/valid-format? "semver"))
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Public
+
+(schema/defn ^:always-validate
+  get-service-version :- SemVerVersion
+  [group-id artifact-id]
+  (let [version (versioneer/get-version group-id artifact-id)]
+    (when-not version
+      (throw (IllegalStateException.
+               (format "Unable to find version number for '%s/%s'"
+                       group-id
+                       artifact-id))))
+    (when-not (semver/valid-format? version)
+      (throw (IllegalStateException.
+               (format "Service '%s/%s' has version that does not comply with semver: '%s'"
+                       group-id
+                       artifact-id
+                       version))))
+    version))
+
 
 (schema/defn service-status-map :- ServiceInfo
   [svc-version status-version status-fn]
--- a/test/puppetlabs/trapperkeeper/services/status/status_core_test.clj
+++ b/test/puppetlabs/trapperkeeper/services/status/status_core_test.clj
@@ -2,9 +2,24 @@
   (:require [clojure.test :refer :all]
             [schema.core :as schema]
             [schema.test :as schema-test]
-            [puppetlabs.trapperkeeper.services.status.status-core :refer :all]))
+            [puppetlabs.trapperkeeper.services.status.status-core :refer :all]
+            [trptcolin.versioneer.core :as versioneer]))
 
 (use-fixtures :once schema-test/validate-schemas)
+
+(deftest get-service-version-test
+  (testing "get-service-version returns a version string that satisfies the schema"
+    ;; This test coverage isn't very thorough, but anything beyond this would
+    ;; really just be testing the underlying libraries that we use to
+    ;; implement it.
+    (is (nil? (schema/check
+                SemVerVersion
+                (get-service-version "puppetlabs" "trapperkeeper-status"))))
+    (is (thrown? IllegalStateException
+                 (get-service-version "fake-group" "artifact-that-does-not-exist")))
+    (with-redefs [versioneer/get-version (constantly "bad-version-string")]
+      (is (thrown? IllegalStateException
+                   (get-service-version "puppetlabs" "trapperkeeper-status"))))))
 
 (deftest update-status-context-test
   (let [status-fns (atom {})]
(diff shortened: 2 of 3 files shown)


src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -30,9 +32,30 @@
...
+(schema/defn ^:always-validate
+  get-service-version :- SemVerVersion
+  [group-id artifact-id]
+  (let [version (versioneer/get-version group-id artifact-id)]
+    (when-not version
+      (throw (IllegalStateException.
+               (format "Unable to find version number for '%s/%s'"
+                       group-id
+                       artifact-id))))
+    (when-not (semver/valid-format? version)
```
MSLilah, line 50: Is the schema validation necessary given that we're doing the same validation manually here?

MSLilah:
Looks good! :+1:

src/puppetlabs/trapperkeeper/services/status/status_core.clj
cprice404, line 50: It's really more for purposes of documentation in my mind.

cprice404:
@rlinehan @KevinCorcoran can you two weigh in on the idea here or in the ticket?  You can see an example usage here: https://github.com/puppetlabs/puppet-server/pull/525/files#diff-f69c98ef06ea1897ed3b68e38fb638d3R31 .  Would be nice to at least get on the same page about how we plan to move forward with this before we write up the docs.

test/puppetlabs/trapperkeeper/services/status/status_core_test.clj
```diff
@@ -2,10 +2,25 @@
...
+  (testing "get-service-version returns a version string that satisfies the schema"
+    ;; This test coverage isn't very thorough, but anything beyond this would
+    ;; really just be testing the underlying libraries that we use to
+    ;; implement it.
+    (is (nil? (schema/check
+                SemVerVersion
+                (get-service-version "puppetlabs" "trapperkeeper-status"))))
+    (is (thrown? IllegalStateException
+                 (get-service-version "fake-group" "artifact-that-does-not-exist")))
+    (with-redefs [versioneer/get-version (constantly "bad-version-string")]
```
KevinCorcoran, line 20: :goberserk:

KevinCorcoran:
I'm :+1: on the idea overall.  I wonder if the function should be named something a bit more general, like `get-artifact-version` ... I don't know how much I like that name, but it seems more correct - there may be multiple TK services in a single project using the status service, after all.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
rlinehan, line 50: If that's the case, then why mark it as `^:always-validate`?

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -30,9 +32,30 @@
...
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Public
 
+(schema/defn ^:always-validate
+  get-service-version :- SemVerVersion
+  [group-id artifact-id]
+  (let [version (versioneer/get-version group-id artifact-id)]
+    (when-not version
+      (throw (IllegalStateException.
```
rlinehan, line 46: I don't think this ever gets triggered, because versioneer returns empty string if it can't find the version.

test/puppetlabs/trapperkeeper/services/status/status_core_test.clj
```diff
@@ -2,10 +2,25 @@
...
 
+(deftest get-service-version-test
+  (testing "get-service-version returns a version string that satisfies the schema"
+    ;; This test coverage isn't very thorough, but anything beyond this would
+    ;; really just be testing the underlying libraries that we use to
+    ;; implement it.
+    (is (nil? (schema/check
+                SemVerVersion
+                (get-service-version "puppetlabs" "trapperkeeper-status"))))
+    (is (thrown? IllegalStateException
```
rlinehan, line 18: I think this doesn't throw the `IllegalStateException` that we're expecting though; the message this has is `Service 'fake-group/artifact-that-does-not-exist' has version that does not comply with semver: ''`.

KevinCorcoran:
Why enforce SemVer?  That seems like it might not play nicely with SNAPSHOTs...

rlinehan:
@KevinCorcoran I tested it (and also the tests currently test it since our current version is a SNAPSHOT), and it works fine with SNAPSHOTS:

``` clojure
user=> (clj-semver.core/version "1.0.0-SNAPSHOT")
{:build nil :major 1 :minor 0 :patch 0 :pre-release "SNAPSHOT"}
```

rlinehan:
I think this function will be really useful, and I'm +1 on this aside from the issue that versioneer returns an empty string rather than nil, so the first error case will never be reached. I kind of agree with @KevinCorcoran about the name of the function, but I don't think that's a huge deal?

KevinCorcoran:
@rlinehan ah, thanks for testing that.

I still don't quite see a reason to enforce SemVer, but not a big deal.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
cprice404, line 50: Honestly, because I don't see any good reason to ever _not_ mark a schemafied function as `:always-validate` until the day we see that causing a performance problem of some kind :)

I can remove it if you all would prefer, I just generally think the cost of those validations is so insignificant compared to anything interesting that our code does, it's never seemed worth spending any energy trying to decide when to use that and when not to.

test/puppetlabs/trapperkeeper/services/status/status_core_test.clj
cprice404, line 20: yeah, I know :)  Any ideas?  I can just ditch that test altogether, but other than that I couldn't think of anything.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
cprice404, line 46: oh, good catch, thanks.

test/puppetlabs/trapperkeeper/services/status/status_core_test.clj
cprice404, line 18: haha :) OK i'll improve the test to have it validate the exception more thoroughly.

cprice404:
> I wonder if the function should be named something a bit more general, like get-artifact-version ... I don't know how much I like that name, but it seems more correct - there may be multiple TK services in a single project using the status service, after all.

Yeah, good call.  I'll change it and can't think of a better name than the one you suggested.

cprice404:
Thanks for reviewing.  Sounds like we've got enough consensus that we can include this in the docs even if it hasn't quite been merged yet.  I'll add a commit to fix up the last few things tomorrow AM.

cprice404:
Closing this in favor of #11 , which is a rebase plus a few extra commits.

PR #9, opened 2015-05-06
(TK-212) Error when service version already exists

MSLilah:
Throw an exception when a user tries to register multiple
callbacks for a service at the same service level.

### DIFF
--- a/test/puppetlabs/trapperkeeper/services/status/status_core_test.clj
+++ b/test/puppetlabs/trapperkeeper/services/status/status_core_test.clj
@@ -15,4 +15,17 @@
 
       (is (nil? (schema/check ServicesInfo @status-fns)))
       (is (= 2 (count (get @status-fns "foo"))))
-      (is (= 1 (count (get @status-fns "bar")))))))
+      (is (= 1 (count (get @status-fns "bar")))))
+
+    (testing (str "registering a service status callback function with a "
+                  "version that already exists causes an error")
+      (is (thrown? IllegalStateException
+                   (update-status-context status-fns "foo"
+                                          "1.1.0" 2 (fn [] "foo repeat")))))
+
+    (testing (str "registering a service status callback function with a "
+                  "status version that already exists but a different service "
+                  "version causes an error")
+      (is (thrown? IllegalStateException
+                   (update-status-context status-fns "foo"
+                                          "1.2.0" 2 (fn [] "foo repeat")))))))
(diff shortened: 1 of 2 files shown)


test/puppetlabs/trapperkeeper/services/status/status_core_test.clj
```diff
@@ -15,4 +15,17 @@
...
+      (is (thrown? IllegalStateException
+                   (update-status-context status-fns "foo"
+                                          "1.1.0" 2 (fn [] "foo repeat")))))
+
+    (testing (str "registering a service status callback function with a "
+                  "status version that already exists but a different service "
+                  "version causes an error")
+      (is (thrown? IllegalStateException
+                   (update-status-context status-fns "foo"
+                                          "1.2.0" 2 (fn [] "foo repeat")))))))
```
cprice404: This raises an interesting question: does it even make sense to allow registration of more than one `service-version` for a service?  I kind of think the answer is "no".  So maybe we should check for that and error out as well?

rlinehan: I would think that we do want to allow registration for more than one `service-version`? For example, we make some changes to puppet server's status function in puppet server 2.5.0 and thus have a puppet server status function v2, but we still want to support puppet server status function v1. Thus we would have in puppet server's init

``` clojure
(register-callback "puppet-server" "2.5.0" 1 puppetserver-status-v1)
(register-callback "puppet-server" "2.5.0" 2 puppetserver-status-v2)
```

rlinehan:
:+1: this looks good to me!

test/puppetlabs/trapperkeeper/services/status/status_core_test.clj
cprice404: We definitely want to support more than one `status-version`, totally agree.  But `service-version`?  That'd be more like if we registered callbacks for `puppet-server` version `2.4.9` and `2.5.0` in the same process... which seems... craycray?  I might be missing something but so far I haven't been able to imagine a situation where that seems like it would make sense.

rlinehan: Ah, I see. I was completely misreading what you were suggesting (I was thinking you were saying not more than one status function registered per service-version, but really what you were saying is no matter how many status functions are registered by a service, they must all have the same service-version). That does seem like it would be really bizarre, and I can see throwing an error for that.

MSLilah: That should be a really easy check, I'll go ahead and add it.

MSLilah:
@rlinehan @cprice404 I've added some error checking for the case in which a user tries to register a callback with a different service version than one that already exists for the same service.

PR #10, opened 2015-05-07
(TK-209) Add :is_running to status service output

rlinehan:
This PR makes the following changes:
- unwrap exceptions from pmap so that they can be handled by our middleware
- add an `:is_running` key to the status service output
- add a `:detail_level` key to the status service output
- add a schema for the `StatusCallbackResponse`
- change hyphens to underscores in JSON output

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/status/status_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_core.clj
@@ -13,11 +13,16 @@
 (def ServiceStatusDetailLevel
   (schema/enum :critical :info :debug))
 
+(def StatusCallbackResponse
+  {:is-running schema/Bool
+   :status schema/Any})
+
 (def ServiceInfo
   {:service-version schema/Str
    :service-status-version schema/Int
-   ;; TODO: specify output of callback function in this schema?
-   :status-fn (schema/make-fn-schema ServiceStatusDetailLevel schema/Any)})
+   ;; Note that while this specifies the input and output for the status
+   ;; function for each service, it does not actually validate these
+   :status-fn (schema/make-fn-schema StatusCallbackResponse ServiceStatusDetailLevel)})
 
 (def ServicesInfo
   {schema/Str [ServiceInfo]})
@@ -25,6 +30,7 @@
 (def ServiceStatus
   {:service-version schema/Str
    :service-status-version schema/Int
+   :is-running (schema/enum true false :unknown)
    :status schema/Any})
 
 (def ServicesStatus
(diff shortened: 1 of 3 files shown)


cprice404:
LGTM, need to add a test to describe the behavior when multiple services are registered and one of them misbehaves, and change the implementation accordingly.

Hmmm... I just noticed that we have a separate ticket for that: TK-175.  Maybe it's OK to punt that work until that ticket comes in, then.

If that's the way we're planning to go then maybe for this one all that's left is to consider consolidating the callback response schema with the service status schema.  Changing that later would be a breaking API change, so it'd be good to make a call on it before we send out our announcement.  (The decision will also affect the documentation.)

Looking good, though!

MSLilah:
I'm :+1: barring the merging of those two schemas. I'm fine with punting work around error handling in the callback functions to a future PR dealing with TK-175.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -13,18 +13,27 @@
...
+;; this is what gets returned in the HTTP response as json, and thus should
+;; use underscores rather than hyphens
 (def ServiceStatus
-  {:service-version schema/Str
-   :service-status-version schema/Int
+  {:service_version schema/Str
+   :service_status_version schema/Int
+   :is_running (schema/enum true false :unknown)
+   :detail_level ServiceStatusDetailLevel
    :status schema/Any})
```
rlinehan, line 42: Ergh, so the problem with reusing `StatusCallbackResponse` in this is that this uses underscores (because ultimately it's going to be JSON output), while that has hyphens. We could just use underscores everywhere, I guess? Or just keep them separate.

rlinehan:
I agree about punting error handling work until TK-175, since I think that ticket is meant for that, and I think we have a little more discussion as to what the right behavior is (although I do agree that it should be in a try/catch).

I've added a commit that makes a schema for the output of `:is-running`, so that it's always a keyword. Unfortunately, I think before we can merge those two schemas we need to have a better way of doing hyphen -> underscore transformation for json output (and I didn't really want to do that at 5pm). However, I think that at this point doing that (and merging these schemas) should not cause any API changes, so hopefully that's not a blocker?

cprice404:
I think you're right that the API should hopefully not be affected when we come back around and deal with the JSON/underscore/combining schemas madness.  Thanks!

cprice404:
@rlinehan do you think it would be worth starting an e-mail thread between jade and the NC team to ask for their advice / warnings re: underscores and JSON?

PR #11, opened 2015-05-08
(TK-210) add get artifact version fn

cprice404:


cprice404:
This is the same as #8 , but with two commits added.  One to address the PR feedback and one to do some whitespace cleanup.  If the whitespace is not to anyone's liking, let me know and I'm happy to change it in a subsequent PR, but if there are no other issues that people find, let's not block the PR on that, so that we can get this in front of other teams.

project.clj
```diff
@@ -9,6 +9,8 @@
 
   :pedantic? :abort
 
+  :exclusions [org.clojure/clojure]
```
KevinCorcoran, line 12: Interesting.  I'm guessing this is needed because either clj-semver or versioneer is stuck on an old version of Clojure?

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -44,31 +46,56 @@
...
+  `artifact-id` should match the maven/leiningen identifiers for the project
+  that the service is defined in."
+  [group-id artifact-id]
+  (let [version (versioneer/get-version group-id artifact-id)]
+    (when (empty? version)
+      (throw (IllegalStateException.
+               (format "Unable to find version number for '%s/%s'"
+                 group-id
+                 artifact-id))))
+    (when-not (semver/valid-format? version)
```
KevinCorcoran, line 90: Why enforce SemVer?

KevinCorcoran:
:+1: aside from question about enforcing SemVer

rlinehan:
:+1: I don't love the "two spaces everywhere" whitespace, but I think that's mostly just because I haven't yet figured out how to get vim to do that. :P

PR #12, opened 2015-05-08
(TK-213) initial docs

cprice404:


documentation/developers.md
```diff
@@ -0,0 +1,39 @@
...
+3. Add `[:StatusService register-status]` to your TK service's deps.
+4. Call `status-core/get-artifact-version` to dynamically get the
+   version info for your service.
+5. Call `register-status` to register a callback function that returns
+   status info for your service.
+6. Define your status callback function.
+
+Code sample:
+
+    (ns foo
```
rlinehan, line 15: Tiny thing: I think we can use ```clojure here to get clojure syntax highlighting, although I think that only works on github.

documentation/query-api.md
```diff
@@ -0,0 +1,213 @@
...
+        ...
+        },
+    ...
+    }
+
+For detailed information, please see the [Wire Format Specification](./wire-formats.md).
+
+NOTE: If any services in your application have registered more than one supported
+status format version, and you are querying at a detail level other than `critical`
+this endpoint will *always* return the latest format.  Therefore,
```
rlinehan, line 64: I don't think currently there is anything in our code that makes it so that `critical` will return all of the registered status outputs? That maybe something we want to add, but that's not currently how it works.

documentation/query-api.md
```diff
@@ -0,0 +1,213 @@
...
+application.
+
+### URL Parameters
+
+* `level`: Optional.  A JSON String from among the legal
+[Status Detail Levels](#StatusDetailLevel) listed above.  Status information for
+the requested service will be provided at the requested level of detail.  If
+not provided, the default level is `"info"`.
+
+* `service-status-version`: Optional.  A JSON integer specifying the desired status
```
rlinehan, line 137: This should be `service_status_version` I think, because we're using underscores in query params.

README.md
```diff
@@ -2,8 +2,16 @@
 
 [![Build Status](https://travis-ci.org/puppetlabs/trapperkeeper-status.svg)](https://travis-ci.org/puppetlabs/trapperkeeper-status)
 
-A trapperkeeper service for getting the status of other trapperkeeper
-services.
+A Trapperkeeper service that provides a web endpoint for getting status
+information about a running Trapperkeeper application.
```
KevinCorcoran, line 6: I think this should pluralized, s/a running Trapperkeeper application/running Trapperkeeper applications

documentation/developers.md
KevinCorcoran, line 15: @rlinehan - do you think this documentation will ever move somewhere besides github?  (I think this is the right place for it; certainly does not seem to belong on docs.pl.com to me).

documentation/query-api.md
```diff
@@ -0,0 +1,213 @@
...
+The response format will be a JSON _Object_, which will look something like this:
+
+    {<service-name>: {
+        "service_version": <service-version>,
+        "service_status_version": <service-status-version>,
+        "detail_level": <detail-level>,
+        "is_running": <service-state>,
+        "status": <any>
+        },
+     <service-name>: {
```
KevinCorcoran, line 54: `$service-name` might be a bit easier to parse that `<service-name>` in this example.  It took my brain a moment to figure out if this was JSON or XML or some frankenstein hybrid of the two.

README.md
cprice404, line 6: I don't think so; there will only be one TK application involved... there may be multiple TK services, but only one application.

documentation/developers.md
cprice404, line 15: There will most definitely be a variant of this on docs.pl.com at some point.  The main goal of this service is to provide Puppet Enterprise users with a uniform way of getting status information about services running in Puppet Enterprise applications.

documentation/query-api.md
```diff
@@ -0,0 +1,213 @@
...
+### Examples
+
+Using `curl` from localhost:
+
+Get the service status for a specified service in the application:
+
+    curl -k https://localhost:8000/status/v1/services/other-service
+
+    {
+        "other-service": {
```
KevinCorcoran, line 166: Should `other-service` be `other_service`?

documentation/query-api.md
cprice404, line 64: Yeah, sorry, that's poor phrasing on my part.  What I've been thinking is that if you specify `critical` as the level, then the `status` field would always be `null`.  (We don't enforce that, and we should probably chat about it to decide whether it's even a good idea, but that was the assumption that I had in my head when writing these docs.)  If that were the case, then theoretically there's no schema left that could vary between format versions, so you wouldn't really need to think about which format version you were requesting / viewing.

This is phrased poorly, regardless, and it probably makes sense to just remove the stanza about 'other than `critical`' for now.

documentation/query-api.md
cprice404, line 137: Yeah, you're right.

documentation/query-api.md
KevinCorcoran, line 166: Yeah, I think all of the keys should use underscores (instead of hyphens) to conform to our style guide.

KevinCorcoran:
Great work, @cprice404!  :gb:   I think this is great overall.

documentation/query-api.md
cprice404, line 54: I was trying to base this on prior art in the PuppetDB docs (which Nick F. has been pretty involved with, historically).  I think I messed up, though, they appear to just use spaces in theirs:

http://docs.puppetlabs.com/puppetdb/2.3/api/query/v4/facts.html

documentation/query-api.md
cprice404, line 166: Consumers can specify whatever they like for the service name, so it gets a bit dicey... but I think you're right that for our examples we should probably just try to make it look consistent :)

cprice404:
Thanks!

documentation/query-api.md
rlinehan, line 166: Except we already have the example of "puppet-server", so if we change this to "other_service" then that's inconsistent... Although if we don't want to change this key to use underscores then that's going to become difficult if we want to do automatic hyphen->underscore transforming.

cprice404, line 166: Well, I don't think we want to change it from what the consumer passes us in the callback registration.  I think the JSON transforming can be restricted to Clojure keywords, hopefully, and leave strings intact.

The "puppet-server" one is still just hypothetical at this point, it only exists in a PR, so if we wanted to set an example that suggests underscores are the way to go, we could just change that PR.

rlinehan, line 166: Yeah, for this PR I just want us to be consistent within the PR.

README.md
KevinCorcoran, line 6: Oh, right

KevinCorcoran:
I'm going to close this since #14 was merged.

PR #13, opened 2015-05-08
(maint) Update error maps to be more inline with convention

rlinehan:
Most other projects at PL seem to have standardized on returning a map with
:kind and :msg keys for errors for http responses. This commit brings the
status service in line with that convention.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/status/ringutils.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/ringutils.clj
@@ -17,7 +17,7 @@
           (catch #(contains? #{:request-data-invalid :service-status-version-not-found}
                              (:type %)) e
             {:status 400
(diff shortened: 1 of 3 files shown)


KevinCorcoran:
These changes look fine, as long as it is decided by the Great Powers That Be that this is, in fact, the correct format to use here.

rlinehan:
It seems like no one has actually decided whether we should use `type` or `kind`, so I'm going to maybe just change this to remove the nesting and keep the keys as it was before, until we have an actual decision on this.

KevinCorcoran:
that works for me.

rlinehan:
updated

KevinCorcoran:
:shipit:

MSLilah:
:shipit:

PR #14, opened 2015-05-08
(TK-213) initial docs

rlinehan:
This PR adds a few commits on top of #12, fixing some links, changing the error format to correspond to #13, and clarifying a note about the `/services` endpoint.


KevinCorcoran:
:+1:

PR #17, opened 2015-05-22
(TK-217) Add status level comparison functions

KevinCorcoran:
Define numeric values for status levels.

Add a level>= function to aid in the implementation of status functions
which return additive information across status levels.

### DIFF
--- a/documentation/developers.md
+++ b/documentation/developers.md
@@ -30,6 +30,33 @@ Code sample:
       v1-status-callback)
     context))
 ```
+
+
+## Implementing Your Status Function
+
+The `puppetlabs.trapperkeeper.services.status.status-core` namespace contains
+some utilities to aid in the implementation of your status functions.  In
+particular, the `level->int` function defines an ordering for status levels as
+`:critical < :info < :debug`, and the `compare-levels` function can be used to 
+compare status levels.  This is especially useful in conjunction with the 
+`cond->` macro from `clojure.core`.  Here's an example of how a status function 
+might be implemented to utilize the `compare-levels` function:
+```clj
+(require '[puppetlabs.trapperkeeper.services.status.status-core :as status-core])
+
+(defn my-status
+  [level]
+  (let [level>= (partial status-core/compare-levels >= level)]
+    {:is-running true
+     :status (cond-> {:this-is-critical "foo"}
+               (level>= :info) (assoc :bar "bar"
+                                      :baz "baz")
+               (level>= :debug) (assoc :x "x"
+                                       :y "y"
+                                       :z "y"))}))
+```
+
+
 ## Details
 
 See [Query API](./query-api.md) and [Wire Format](./wire-formats.md) for details
--- a/src/puppetlabs/trapperkeeper/services/status/status_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_core.clj
@@ -95,6 +95,21 @@
                  version))))
     version))
 
+(defn level->int
+  "Returns an integer which represents the given status level.
+   The ordering of levels is :critical < :info < :debug."
+  [level]
+  (case level
+    :critical 0
+    :info 1
+    :debug 2))
+
+(defn compare-levels
+  "Converts the two status levels to integers using level->int and then
+   invokes f, passing the two integers as arguments.  Especially useful for
+   comparing two status levels."
+  [f level1 level2]
+  (f (level->int level1) (level->int level2)))
 
 (schema/defn service-status-map :- ServiceInfo
   [svc-version status-version status-fn]
--- a/src/puppetlabs/trapperkeeper/services/status/status_service.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_service.clj
@@ -8,7 +8,11 @@
 (defprotocol StatusService
   (register-status [this service-name service-version status-version status-fn]
     "Register a status callback function for a service by adding it to the
-    status service context."))
+    status service context.  status-fn must be a function of arity 1 which takes
+    the status level as a keyword and returns the status information for
+    the given level.  The return value of the callback function must satisfy
+    the puppetlabs.trapperkeeper.services.status.status-core/StatusCallbackResponse
+    schema."))
 
 (defservice status-service
   StatusService
--- a/test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
+++ b/test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
@@ -3,9 +3,10 @@
             [cheshire.core :as json]
             [schema.test :as schema-test]
             [puppetlabs.http.client.sync :as http-client]
-            [puppetlabs.trapperkeeper.core :refer [defservice]]
+            [puppetlabs.trapperkeeper.core :refer [defservice service]]
             [puppetlabs.trapperkeeper.testutils.bootstrap :refer :all]
-            [puppetlabs.trapperkeeper.services.status.status-service :refer :all]
+            [puppetlabs.trapperkeeper.services.status.status-service :refer [status-service]]
+            [puppetlabs.trapperkeeper.services.status.status-core :as status-core]
             [puppetlabs.trapperkeeper.services.webrouting.webrouting-service :as webrouting-service]
             [puppetlabs.trapperkeeper.services.webserver.jetty9-service :as jetty9-service]))
 
@@ -14,8 +15,7 @@
 (def status-service-config
   {:webserver {:port 8180
                :host "0.0.0.0"}
-   :web-router-service
-   {:puppetlabs.trapperkeeper.services.status.status-service/status-service "/status"}})
+   :web-router-service {:puppetlabs.trapperkeeper.services.status.status-service/status-service "/status"}})
 
 (defmacro with-status-service
   "Macro to start the status service and its dependencies (jetty9 and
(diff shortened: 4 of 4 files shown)


test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -176,3 +176,46 @@
...
+  (testing "use of level>= to implement a status function"
+    (let [my-status (fn [level]
+                      (let [level>= (partial status-core/level>= level)]
+                        {:is-running true
+                         :status (cond-> {:this-is-critical "foo"}
+                                   (level>= :info) (assoc :bar "bar"
+                                                          :baz "baz")
+                                   (level>= :debug) (assoc :x "x"
+                                                           :y "y"
+                                                           :z "y"))}))
```
KevinCorcoran, line 194: This is the interesting bit of this PR - how would we like the implementation of this status function to look?  This isn't _exactly_ what's described in TK-217 (although my reading of that ticket was that it was sort of a rough idea, and that what I came up with here is what @cprice404 was going forward in his description).

I am reasonably happy with how this looks, but would be happy to consider alternative - @cprice404, especially if this isn't exactly what you had in mind.

KevinCorcoran:
Oh, and this'll need some docs, if we like it.

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
cprice404, line 194: I'm OK with this; it's better than what we had before and it's better than what I had in my pseudo-code.

That said, it might be nice to separate out the ">=" logic so that the user could compose whatever fns they wanted?

So, your partial could look like this:

``` clj
(let [level>= (partial >= (status-core/numeric-level level)]
```

That would be a bit more flexible, and from the user perspective the rest of your example code could stay exactly as it is... and we could include an example like yours in the docs?  If you are on board with that, I'm not necessarily in love with the name `numeric-level`... could be like `level-value` or something.

cprice404, line 194: Also we should add a `condp->` threading macro to kitchensink :)

KevinCorcoran, line 194: I think you're on to something there ... abstracting-out the comparison seems like a good idea.  However, I like the convenience of having a pre-defined `level>=` function - and it seems like similar functions like `level<=` and `level=` functions would also be useful - and all 3 of these could be implemented on top of something like ...

``` clj
(defn compare-levels
  [operator level1 level2]
  (operator (level->numeric level1) (level->numeric level2))
```

.. whacha think of that?

Switching topics to one of the 2 Hard Problems of Computer Science ...

Sure, I'm not married to `level->numeric`, although it seems a bit clearer than `level-value` ... how about `numeric-value` or `numeric-level-value`?

cprice404, line 194: Yeah, don't really love nor hate any of those.  `level->numeric` or `level->int` or `level-as-int`, I'm probably ok with any of those.  I think maybe `level->int` is my favorite.

cprice404:
+1 other than discussion about naming and providing a fn that doesn't implicitly assume the comparison operator.

KevinCorcoran:
I've updated this PR with a couple new commits:
- added docs
- renamed `level->numeric` to `level->int`
- add additional comparison functions
- add function which takes an arbitrary function (like a comparison function) as a parameter

I think this is ready for merge, but happy to discuss further if there is any more feedback.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -95,6 +95,40 @@
...
+(defn level->int
+  "Returns an integer which represents the given status level.
+   The ordering of levels is :critical < :info < :debug."
+  [level]
+  (case level
+    :critical 0
+    :info 1
+    :debug 2))
+
+(defn apply-to-int-levels
```
cprice404: I think I would just call this `compare-levels`, maybe?

documentation/developers.md
```diff
@@ -30,6 +30,32 @@ Code sample:
...
+  [level]
+  (let [level>= (partial status-core/level>= level)]
+    {:is-running true
+     :status (cond-> {:this-is-critical "foo"}
+               (level>= :info) (assoc :bar "bar"
+                                      :baz "baz")
+               (level>= :debug) (assoc :x "x"
+                                       :y "y"
+                                       :z "y"))}))
+```
```
cprice404, line 57: Do you think maybe it'd be worth adding another example here that shows how to do it the "manual" way if you opt out of the utility comparator fns?  e.g.:

``` clj
(defn my-status
   [level]
   (let [level>= (partial status-core/compare-level >= level)]
      (cond-> {...}
          ...
       )))
```

cprice404, line 57: Actually, looking at that snippet above, it kind of seems like there is almost never going to be a use case where someone doesn't want to close over a partial function in order to use the `cond->` macro, and if people are already going to be defining that partial function themselves anyway, I'm going back to where I was before where I'm kind of wondering whether or not it's even valuable to provide those utility fns that bake in the comparator.  They don't reduce the lines of code for the consumer, and we agreed we need the generic one anyway, so it seems like they just adds extra LOC and tests to the status library.

If we do include them I'll probably be inclined to use `compare-level` directly in my own consumer code anyway, because it just seems more readable to me.  Since clojure uses prefix notation for everything, `(compare-level >= :info :critical)` seems fairly intuitive to me since the comparison operator is in the "normal" position relative to the two arguments being compared, and it's nice to see that explicitly in my own code.  Whereas with the `level>=` fn, it kind of reads like `(level >= ...)` to me, which then sort of makes the ordering of the two arguments that are being compared ambiguous?

I dunno, maybe that's just me.  I guess it's pretty harmless to leave those extra fns in, but looking at the two code snippets above, I'm pretty sure I'll end up just using the `compare-level` fn directly.

cprice404, line 57: One other thought is that it's always easier to add API in the future than it is to remove it...

cprice404:
So, this looks great to me.  I have the two outstanding suggestions:
- Change the name of `apply-to-int-levels` to `compare-levels`.
- Possibly remove the built-in >= fns, though if I'm on my own on that one, I don't have a problem with those coming in.

Would be great if @rlinehan or @fpringvaldsen could review today, and then whatever the consensus is, let's just get this merged and do a release.  I'd very much like to be able to merge up another app tomorrow (Thurs) that will need to have a dependency on a release that contains this functionality.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
KevinCorcoran: That was my original thought as well, as that's how we intended it to be used, but what it actually does it more general than just a comparison, so that name doesn't seem quite right to me.

documentation/developers.md
KevinCorcoran, line 57: As Joshua Bloch says, "when in doubt, leave it out".

src/puppetlabs/trapperkeeper/services/status/status_core.clj
KevinCorcoran: That said, I don't like this name very much and would be happy to change it to something else - I just think it needs to accurately reflect the behavior.

rlinehan: Is there something other than comparing that you would likely use this for though? If we can't come up with a better name, I would go with `compare-levels` over `apply-to-int-levels`, because I very hard time parsing the latter and understanding what it means.

documentation/developers.md
rlinehan, line 57: I agree with @cprice404, and I think it would be easier to add these back in if their needed down the line, but right now it doesn't seem like their really needed.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
KevinCorcoran: @rlinehan and I spoke IRL and we decided to change this to `compare-levels`.

KevinCorcoran:
@rlinehan @cprice404 I've made the changes we discussed.  I've also squashed the commits into one, since things were getting a bit messy.

src/puppetlabs/trapperkeeper/services/status/status_service.clj
```diff
@@ -8,7 +8,11 @@
 (defprotocol StatusService
   (register-status [this service-name service-version status-version status-fn]
     "Register a status callback function for a service by adding it to the
-    status service context."))
+    status service context.  status-fn must be a function of arity 1 which takes
+    the status level as a keyword and returns the status information for
+    the given level.  The return value of the callback function must satisfy
+    the puppetlabs.trapperkeeper.services.status.status-core/StatusCallbackResponse
```
rlinehan, line 14: IIRC, there's currently nothing enforcing that this must satisfy this schema, but I think we were maybe going to make that the case in another ticket?

rlinehan:
:+1: :shipit:

src/puppetlabs/trapperkeeper/services/status/status_service.clj
KevinCorcoran, line 14: Right, I guess it's not strictly enforced, but status functions still need to conform to it (i.e. `:is-running` should most likely be there, and the rest of the data needs to be under `:status` for it to end up in the response).  Perhaps I could have explained that a bit better in this docstring.

PR #18, opened 2015-05-27
Add status.check namespace with disk & db check.

aperiodic:
Add the `puppetlabs.trapperkeeper.services.status.check` namespace,
which contains shared status check functions for e.g. the database or
the disk.

Add `db-up?`, which takes a database connection spec and runs a purely
arithmetic query (which will succeed no matter the schema) and checks
the result; if it matches the expected result, the check returns `true`.
If the result doesn't match, or any exceptions are thrown in this
process, the check returns `false`

Add `disk-writable?`, which is a zero-arity function that attempts to
write random characters to a temporary file, reads the file back in, and
check that what was read matches what was written. If so, the check
returns `true`; if the match fails or any exceptions were thrown in this
process, the check returns `false`. The temporary file is deleted with
a `finally` clause.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -12,7 +12,9 @@
   :exclusions [org.clojure/clojure]
 
   :dependencies [[org.clojure/clojure "1.6.0"]
+                 [org.clojure/java.jdbc "0.3.7"]
                  [cheshire "5.3.1"]
+                 [org.postgresql/postgresql "9.3-1100-jdbc41"]
                  [prismatic/schema "0.4.0"]
                  [ring/ring-json "0.3.1" :exclusions [ring/ring-core]]
                  [ring/ring-defaults "0.1.5"]
--- a/src/puppetlabs/trapperkeeper/services/status/check.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/check.clj
@@ -1,0 +1,28 @@
+(ns puppetlabs.trapperkeeper.services.status.check
+  "Shared status check functions."
+  (:require [clojure.java.jdbc :as sql])
+  (:import [java.io File IOException]))
+
+(defn db-up?
+  [db-spec]
+  (try (let [select-42 "SELECT (a - b) AS answer FROM (VALUES ((7 * 7), 7)) AS x(a, b)"
+             [{:keys [answer]}] (sql/query db-spec [select-42])]
+         (= answer 42))
+    (catch Exception _
+      false)))
+
+(defn disk-writable?
+  []
+  (if-let [temp-file (try (File/createTempFile "tk-disk-check" ".txt")
+                       (catch IOException _ nil))]
+    (let [payload (apply str (map char (repeatedly 4095 #(rand-nth (range 97 123)))))]
+      (try
+        (spit temp-file payload)
+        (= (slurp temp-file) payload)
+        (catch Exception _
+          false)
+        (finally
+          (try (.delete temp-file)
+            (catch IOException _
+              false)))))
+    false))
--- a/test/puppetlabs/trapperkeeper/services/status/check_test.clj
+++ b/test/puppetlabs/trapperkeeper/services/status/check_test.clj
@@ -1,0 +1,57 @@
+(ns puppetlabs.trapperkeeper.services.status.check-test
+  (:require [clojure.java.jdbc :as sql]
+            [clojure.test :refer :all]
+            [puppetlabs.trapperkeeper.services.status.check :refer :all]))
+
+(deftest db-up?-test
+  (testing "db-up?"
+    (testing "connects to a database, runs an arithmetic query, and checks the result"
+      (let [!query (atom nil)
+            !db (atom nil)
+            db-spec {:conninfo "postgresql://user:pass@localhost:5432/db"}]
+        (with-redefs [sql/query (fn [db sql-vec]
+                                  (reset! !db db)
+                                  (reset! !query sql-vec)
+                                  [{:answer 42}])]
+          (is (db-up? db-spec))
+          (is (= db-spec @!db))
+          (is (= ["SELECT (a - b) AS answer FROM (VALUES ((7 * 7), 7)) AS x(a, b)"]
+                 @!query)))))
+
+    (testing "returns false if the query function throws an exception"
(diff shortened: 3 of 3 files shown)


src/puppetlabs/trapperkeeper/services/status/check.clj
```diff
@@ -0,0 +1,26 @@
...
+  [db-spec]
+  (try (let [select-42 "SELECT (a - b) AS answer FROM (VALUES ((7 * 7), 7)) AS x(a, b)"
+             [{:keys [answer]}] (sql/query db-spec [select-42])]
+         (= answer 42))
+    (catch Exception _
+      false)))
+
+(defn disk-writable?
+  []
+  (let [temp-file (File/createTempFile "tk-disk-check" ".txt")
```
juxtin: If the disk is unwritable, won't this step fail?

vilmibm: yeah, java docs say it will throw "IOException - If a file could not be created"

aperiodic: derp

aperiodic:
@holguinj @nathanielksmith I've updated this to handle a failure to create the temp file in the disk-writable? check, and also to remove the merge conflicts.

juxtin:
I'm :+1: on this. @nathanielksmith ?

project.clj
```diff
@@ -12,7 +12,9 @@
   :exclusions [org.clojure/clojure]
 
   :dependencies [[org.clojure/clojure "1.6.0"]
+                 [org.clojure/java.jdbc "0.3.7"]
                  [cheshire "5.3.1"]
+                 [org.postgresql/postgresql "9.3-1100-jdbc41"]
```
rlinehan, line 17: It looks like this dependency is never used.

aperiodic, line 17: It's not directly used. What it does is provide a postgresql JDBC driver for clojure.java.jdbc. Without it, clojure.java.jdbc is unable to connect to a postgres database.

aperiodic, line 17: Although I suppose we could expect any consuming service to rely on the JDBC drivers that they need?

KevinCorcoran, line 17: Yeah, I think that is the correct expectation here, and that this dependency should be removed.

juxtin, line 17: just make it a dev dependency?

KevinCorcoran:
The additional functionality in this PR doesn't seem like it fits into this repo.  This may be a case me of not fully understanding the goals/responsibilities/purview of the status service, as laid out by @cprice404 and @pcarlisle, but putting anything specific to JDBC, postgres, or databases at all into this repo seems incorrect to me.  

My understanding is that the status service should generally apply to any TK service which already uses tk-jetty as well.  Putting DB-specific stuff in here just seems like the wrong place, as it pulls in an extra dependency, and makes part of the library non-applicable to many TK services (those which don't use JDBC).

If it's decided that we need to expand the scope of this library to be a bit more inclusive and include things like this DB-specific functionality, that's okay with me, I suppose ... I just feel like there's a conversation to be had there before we do so, because it seems like additional scope beyond what's been discussed - at least, AFAIK - and I think @rlinehan has a similar perspective on the matter.

aperiodic:
@KevinCorcoran I agree that the postgres JDBC driver shouldn't have been included. However, I do think it makes sense to include these functions with this library. Why make every service reinvent the weel in order to check if their database is up and the disk is writable? I don't think pulling in a small dependency (23 KB) that some services might not use outweighs the benefit of code reuse in this case.

KevinCorcoran:
I'm not suggesting that every service needs to reinvent the wheel; having this code in shared library makes sense to me, I'm just not sure that this is the correct library.  

It's not the size of the dependency's .jar file that is a concern.  In general, I'm in favor of libraries having minimal dependencies, but that's not actually what I think is really important in this case - my point was more about separation of concerns, and whether or not the added code really made sense in this project, or perhaps a separate library.

aperiodic:
From the README, my conception of the purpose of this library is to assist other services in reporting their status over HTTP/S. To that end, providing functions that the consuming services can use in order to determine their status seems appropriate

If we created a separate library just for these status checking functions, then as far as console services is concerned, every project that depends on the TK status library would also depend on this new library. From that perspective, it seems silly to have two separate libraries, especially considering the overhead of setting up CI and whatnot for the new library.

cprice404:
@aperiodic we already have a `jdbc-utils` library that all of the db-related projects should be depending on, correct?  Why not put the db-specific stuff there?

src/puppetlabs/trapperkeeper/services/status/check.clj
```diff
@@ -0,0 +1,28 @@
...
+(defn db-up?
+  [db-spec]
+  (try (let [select-42 "SELECT (a - b) AS answer FROM (VALUES ((7 * 7), 7)) AS x(a, b)"
+             [{:keys [answer]}] (sql/query db-spec [select-42])]
+         (= answer 42))
+    (catch Exception _
+      false)))
+
+(defn disk-writable?
+  []
```
cprice404, line 15: It seems like this would be better if it took a path argument... how often is the temp directory really the thing that you care about writing to?  And isn't it possible that the temp dir is writable and the real dir isn't?  They may not even be on the same disk partition as one another?

Do you plan on calling this every time a request to the status endpoint is made?  In the case of a load balancer hitting the endpoint very frequently, seems like the extra disk IO (though admittedly tiny) would be undesirable?  If the intent is to only hit it for levels like `:debug`, maybe we should put that in the docstring?

cprice404:
I agree with @KevinCorcoran that the db stuff seems pretty out-of-place here.  IMO it's a decent idea to have some generic `check` functions that could be meaningful across services, but we should be careful about the scope and API of said functions... I'd love to see the db utilities moved to the `jdbc-utils` library and the disk fn modified to require a path argument before we release these changes, but that's just my 2 cents...

src/puppetlabs/trapperkeeper/services/status/check.clj
aperiodic, line 15: Those are good points, especially about the drawbacks of only using the temporary directory.

aperiodic:
@KevinCorcoran @cprice404 I've opened PR #21 to remove the DB check and update the disk check function.

PR #20, opened 2015-05-29
(maint) Remove postgresql dependency.

aperiodic:
As downstream consumers that want to use the database status check will
supply their own JDBC drivers, there's no need to include this
dependency.


KevinCorcoran:
:+1:

PR #21, opened 2015-06-02
(maint) No DB, disk-writable? takes path

aperiodic:
Remove the `db-up?` status check, its tests, the
org.clojure/java.jdbc dependency, and the postgresql JDBC drivers
dependency.

Change the `disk-writable?` check to take a directory argument, and
create the file it writes to & reads from inside that directory, rather
than in the system's temporary directory.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/status/check.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/check.clj
@@ -1,28 +1,30 @@
 (ns puppetlabs.trapperkeeper.services.status.check
   "Shared status check functions."
-  (:require [clojure.java.jdbc :as sql])
   (:import [java.io File IOException]))
 
-(defn db-up?
-  [db-spec]
-  (try (let [select-42 "SELECT (a - b) AS answer FROM (VALUES ((7 * 7), 7)) AS x(a, b)"
-             [{:keys [answer]}] (sql/query db-spec [select-42])]
-         (= answer 42))
-    (catch Exception _
-      false)))
-
 (defn disk-writable?
-  []
-  (if-let [temp-file (try (File/createTempFile "tk-disk-check" ".txt")
-                       (catch IOException _ nil))]
-    (let [payload (apply str (map char (repeatedly 4095 #(rand-nth (range 97 123)))))]
-      (try
-        (spit temp-file payload)
-        (= (slurp temp-file) payload)
-        (catch Exception _
-          false)
-        (finally
-          (try (.delete temp-file)
-            (catch IOException _
-              false)))))
-    false))
+  "Given a directory, check whether the FS that the directory is in seems
+  healthy by writing 4k random letters to a file in that directory, and then
+  read them back in. If what was read matches what was written, then return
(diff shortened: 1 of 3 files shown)


cprice404:
+1

src/puppetlabs/trapperkeeper/services/status/check.clj
```diff
@@ -1,28 +1,30 @@
...
+  "Given a directory, check whether the FS that the directory is in seems
+  healthy by writing 4k random letters to a file in that directory, and then
+  read them back in. If what was read matches what was written, then return
+  true; if there's a mismatch or anything goes wrong in the process, return
+  false.
+
+  Since this check involves writing to disk, it may not be appropriate to put at
+  a high status check level like :critical or :info."
+  [directory]
+  (let [rand-letter #(char (rand-nth (range 97 123)))
```
KevinCorcoran, line 15: :hamster:

src/puppetlabs/trapperkeeper/services/status/check.clj
```diff
@@ -1,28 +1,30 @@
...
+        fname (str "tk-disk-check-" (apply str (repeatedly 10 rand-letter)) ".txt")]
+    (if-let [file (File. directory fname)]
```
KevinCorcoran, line 17: `if-let` can be replaced with just `let` here, right?  There is no way the `File` constructor returns something falsey.

src/puppetlabs/trapperkeeper/services/status/check.clj
```diff
@@ -1,28 +1,30 @@
...
+      (let [payload (apply str (repeatedly 4095 rand-letter))]
+        (try
+          (when (.exists file)
+            (.delete file))
+          (spit file payload)
+          (= (slurp file) payload)
```
KevinCorcoran, line 23: The call to `=` here has no effect, just returns true or false ... were you thinking it threw an exception?

test/puppetlabs/trapperkeeper/services/status/check_test.clj
```diff
@@ -1,57 +1,34 @@
...
-                      slurp (fn [file]
-                              (is (= @!file file))
-                              @!content)]
-          (is (disk-writable?))
-          (is (not (.exists @!file))))))
+  (let [temp-dir (System/getProperty "java.io.tmpdir")]
+    (testing "disk-writable?"
+      (testing "writes to a temp file, checks what was written, and deletes the file."
+        (let [!file (atom nil)
+              !content (atom nil)]
```
KevinCorcoran, line 10: ooh, I like this naming convention, I've been wanting something like this for atoms.

src/puppetlabs/trapperkeeper/services/status/check.clj
KevinCorcoran, line 23: Oh my, it's early in the morning.  I did not realize this was the last expression before `catch`.  :coffee:

KevinCorcoran:
:+1: Thanks for taking care of this @aperiodic

PR #22, opened 2015-06-12
(TK-176) HTTPS proxy for the status service

jpinsonault:
This almost definitely is just a first pass, and I'm sure there are many things that can be improved.

The `status-proxy-service` is configured with a few options in the `:status-proxy` key in a config file. The options are `target-host`, `target-port`, `target-url`, and any ssl configuration can be specified in the `target-options` key.

I lifted the SSL certs for the tests from the tk-jetty9 repo.

The tests verify that the proxy functions correctly, and do some basic checking to make sure the proxy doesn't proxy anything it shouldn't. Since security is such an important concern, and the premise of the `status-proxy-service` already means that you're opening a (tiny) hole in your security for convenience, I'd be interested in any better ideas on how to ensure that the proxy isn't opening any unintended security holes.

Also, I'm new to TK, so pile on the criticism and advice.


jpinsonault:
Ping @cprice404 @camlow325 for review, or anyone else you think would want to weigh in

MSLilah:
Since the `status-service` and `status-proxy-service` are two separate services, I'd recommend putting the `status-proxy-service` in its own file. Likewise, I'd recommend putting the tests for it into their own file as well.

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -219,3 +240,83 @@
                     :y "y"
                     :z "y"}
                   (response->status resp)))))))))
+
+(deftest proxy-ssl-status-endpoint-test
```
MSLilah: I played around with running these tests a bit, and this particular test isn't being run. It's also not being registered as a test (if I comment it out or remove it, the number of tests/assertions run by leiningen doesn't decrease when I run `lein test`). 

Moving it to the bottom of the file causes the assertions to be run, but again, if I comment out the test the number of tests/assertions run by leiningen does not decrease. I have no idea why this is happening but it seems really weird.

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -219,3 +240,83 @@
...
+      ssl-status-service-config
+      ; Start the proxy service
+      (with-app-with-config
+        proxy-app
+        [jetty9-service/jetty9-service
+         webrouting-service/webrouting-service
+         status-proxy-service]
+        status-proxy-service-config
+        ; Make HTTP request to the proxy, which will forward it as an HTTPS request
+        (let [resp (http-client/get "http://localhost:8181/status-proxy/v1/services")
```
MSLilah: I think we should also querying for a specific service to ensure that that is being proxied.

Might also be worth doing a simple test to ensure that the `level` or `service_status_version` query params work with the proxy service.

MSLilah:
:chart_with_upwards_trend: This is looking good! Mostly I had some suggestions for additional test coverage and on some code reorganization, and I found an issue with leiningen not running one of the `deftest` blocks

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
jpinsonault: I think I defined the same test twice

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
jpinsonault: Yeah, sounds like a good idea

jpinsonault:
@fpringvaldsen Thanks for the feedback, I moved the status service and tests to separate files, added a test for a specific service and one using query params. I renamed the second test, which hopefully solves your lein issue

src/puppetlabs/trapperkeeper/services/status/status_proxy_service.clj
```diff
@@ -0,0 +1,16 @@
...
+  (:require [clojure.tools.logging :as log]
+    [puppetlabs.trapperkeeper.core :refer [defservice]]))
+
+(defservice status-proxy-service
+  [[:WebroutingService add-proxy-route]
+   [:ConfigService get-in-config]]
+  (init [this context]
+    (add-proxy-route
+      this
+      {:host (get-in-config [:status-proxy :target-host])
```
camlow325: Rather than just exposing this in external configuration as essentially a serialized form of the arguments that the `add-proxy-route` function takes, I think it might be better to model the external configuration on an existing service that has proxy support.  For example, this could look like the "proxy-config" blob that the `reverse-proxy-ca-service` uses - https://github.com/puppetlabs/pe-puppet-server-extensions/blob/master/documentation/config_settings.markdown#caconf.  By doing it this way, we might be able to more easily migrate to a different underlying representation of the proxy mechanism later / abstract away some detailed configuration options in `add-proxy-route` that the `status-proxy-service` wouldn't need to use.

src/puppetlabs/trapperkeeper/services/status/status_service.clj
```diff
@@ -32,3 +32,16 @@
     (log/infof "Registering status callback function for %s service" service-name)
     (core/update-status-context (:status-fns (service-context this))
       service-name service-version status-version status-fn)))
+
+(defservice status-proxy-service
```
camlow325: Is this duplicated intentionally from `status_proxy_service.clj`?  This doesn't seem to be needed.

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -5,7 +5,7 @@
             [puppetlabs.http.client.sync :as http-client]
             [puppetlabs.trapperkeeper.core :refer [defservice service]]
             [puppetlabs.trapperkeeper.testutils.bootstrap :refer :all]
-            [puppetlabs.trapperkeeper.services.status.status-service :refer [status-service]]
+            [puppetlabs.trapperkeeper.services.status.status-service :refer [status-service status-proxy-service]]
```
MSLilah: The `status-proxy-service` shouldn't be needed in this test file.

camlow325:
:see_no_evil: Looks fine to me.  Just had the one question about whether we want to expose the full set of `add_proxy_route` configurability through the configuration in this service or if we just want a constrained subset, possibly modeled on the configuration that an existing service like the `reverse-proxy-ca` uses.  Also noticed that the service definition seems to have been duplicated across two namespaces.

My larger question, noted in comments in the JIRA ticket, is whether we need to do this work at all - i.e., whether it would be feasible to not have a "status-proxy" service and just allow the "real" status service to be configured on multiple ports/servers within the same Java process - including one non-encrypted one that the user would configure - and just rely upon the fact that the status service would be able to fulfill any registered callbacks within that process.

src/puppetlabs/trapperkeeper/services/status/status_service.clj
jpinsonault: Yeah, good catch. Definitely thought I'd removed it from the file.

src/puppetlabs/trapperkeeper/services/status/status_proxy_service.clj
jpinsonault: This sounds good to me. And the `:scheme :https` bit exposes the implementation, which would make it difficult to change the config later too.

jpinsonault:
@camlow325 I'll work on your feedback, and let @cprice404 weigh in on your question.

src/puppetlabs/trapperkeeper/services/status/status_proxy_service.clj
cprice404: +1, modeling the config on the existing CA proxy config is a great suggestion.

cprice404:
@camlow325 I will respond to that on the ticket.

jpinsonault:
@camlow325 I switched to the simpler configuration format and took out the extra code left lying around.
@fpringvaldsen Removed the unneeded reference to `status-proxy-service`

dev-resources/config/ssl/certs/ca.pem
```diff
@@ -0,0 +1,30 @@
...
++ENTML76+oq1d7XTqAh8mnSJOIc6eMUQqBBYl/5BYa4IzkvkrS4ai4vg0ihKr6pf
+C8XVINTjxChqK08dEr690sPD/3DwPPVGqG+qIyIv6u5buVLniLWiaq4HOG54i/yH
+k21W6A5UizRxb5GwVxWfjsHcLr2pvPq/ipRP/sCNBgKdziaGQm+IqeJiI6bQegx3
+ZF5UX2CaIkZsEM0v2yQ7/t9rSCfpd/eEYMdcWcnSfEf/OOA/ti7jnpapENXturLo
+6/TNJarHYQlkpEsaRfiiSmomAn3TNOOuqBhWtj0fdy21/fETxVVzPp/CL7EvkUnU
+26SguZdBkGf22yZViKRq22eDnM/frsoSMTkIbkayDQ8Hx9JVpfPmdJWQyL86etsd
+uCR5OXJtd7vxRZT15m2cFvdpRW3VZbqFyIwe0NgJUs3FRrjFZdqQiBsvWwQjK8jN
+A3+IAAz4pk0A4xpQvNwvUzo4wyQB8/PTcmZoPBzeDaJScqSto2r3kvOSqvm1PSc6
+gneGkbPvQpQRH9HVxzaEEcCrcZYFZXpCkF0ACuB+smgfMVY2Sno=
+-----END CERTIFICATE-----
```
cprice404: We've been trying to start being more consistent about namespacing test resources in a directory structure that matches the test that is consuming them.  Have a look at the layout of the `dev-resources` dir in puppetserver for an example.

src/puppetlabs/trapperkeeper/services/status/status_proxy_service.clj
```diff
@@ -0,0 +1,22 @@
...
+          port (.getPort target-url)
+          path (.getPath target-url)]
+      (add-proxy-route
+        this
+        {:host host
+         :port port
+         :path path}
+        {:ssl-config (get-in-config [:status-proxy :ssl-opts])
+         :scheme     :https}))
+    (log/info "Initializing status service proxy")
```
cprice404: Nitpick but it might be worth putting this log message _above_ the other initialization logic, so that if a failure occurs, the log message will appear in the logs just before the error.

cprice404:
@jpinsonault I had a few minor comments but I'm mostly +1 on the code.

We are going to need some non-trivial documentation about this stuff; I'd be OK with capturing that in a separate ticket so as not to block this PR, but if we're going to do that I think we should create a ticket for it now and make sure it comes in in the next sprint (if not sooner).  Might be worth adding a comment referencing the new ticket number as a TODO in the code somewhere.

cprice404:
or if you prefer to just hack on the docs over the rest of this sprint and include them as part of this PR, that'd be fine too.  I can give you some guidelines on what the docs need to cover if you are looking for some direction.

jpinsonault:
@cprice404 Might as well get the documentation done now rather than later. Guidelines would be great

dev-resources/config/ssl/certs/ca.pem
jpinsonault: Cool, I'll adjust the directories

jpinsonault:
@cprice404 Namespaced the path in `/dev-resources` and moved the logging message above the initialization logic

src/puppetlabs/trapperkeeper/services/status/status_proxy_service.clj
```diff
@@ -0,0 +1,22 @@
...
+          host (.getHost target-url)
+          port (.getPort target-url)
+          path (.getPath target-url)]
+      (add-proxy-route
+        this
+        {:host host
+         :port port
+         :path path}
+        {:ssl-config (get-in-config [:status-proxy :ssl-opts])
+         :scheme     :https}))
```
camlow325, line 21: Not that anyone's likely to use it this for anything but proxying to an "https" URL, but maybe it would be good to do some sort of validation on this scheme.  Like by doing a `.getProtocol` call on the `target-url` and translate a return value of `http` or `https` to the corresponding keyword else fail startup with some sort of "unsupported scheme" error.

src/puppetlabs/trapperkeeper/services/status/status_proxy_service.clj
```diff
@@ -0,0 +1,22 @@
...
+    (let [target-url (URL. (get-in-config [:status-proxy :proxy-target-url]))
+          host (.getHost target-url)
+          port (.getPort target-url)
+          path (.getPath target-url)]
+      (add-proxy-route
+        this
+        {:host host
+         :port port
+         :path path}
+        {:ssl-config (get-in-config [:status-proxy :ssl-opts])
```
camlow325, line 20: If `ssl-opts` is not supplied in the configuration, what does this do?  Will `add-proxy-route` effectively use `:use-server-config` (which is probably a reasonable thing to do) or will it blow up?  Looks like from the schema validation that tk-jetty9 does that it might blow up - https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/1.3.1/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj#L82-L84.

camlow325, line 20: Hmmm, actually defaulting to `:use-server-config` probably isn't going to make sense for most uses of the service anyway since the status proxy would typically be registered with a Jetty server instance which probably wouldn't have any SSL configuration associated with it because it would only have a plaintext port configured for it.

Maybe it would be better to do some validation in here to confirm that the `ssl-opts` is non-empty and matches an expected schema rather than deferring to Jetty to validate this upstream and produce a potentially less useful failure message, though.

cprice404, line 20: I agree, though would probably be willing to let improved error handling spill over to a follow-up ticket.  I think I'd prioritize docs over that.

cprice404:
@jpinsonault I will add some notes to the ticket re: docs.

src/puppetlabs/trapperkeeper/services/status/status_proxy_service.clj
jpinsonault, line 20: Agreed to all of this. I'll create a new ticket for it

camlow325, line 20: :+1: for a separate ticket to cover the extra error handling work.

jpinsonault:
@cprice404 I added some documentation that I think covers the major points. It would be cool to use the configs puppetserver actually uses, but the status service isn't in a build yet, right? So I just used a very simple tk-jetty configuration

KevinCorcoran:
I reviewed the docs added in 5f838b64f661e1bc86838ee54c438bb79889de17 - LGTM.  :+1:

MSLilah:
@jpinsonault I just had a few minor comments on the docs, after that I think it'll be ready to merge.

I think the only open question left is @camlow325's question on the scheme validation. I think we could get away with doing that in a separate PR, so if it's okay with @jpinsonault and @KevinCorcoran I'll go ahead and merge this once my feedback is addressed.

jpinsonault:
@fpringvaldsen Updated the doc. I also made TK-245 to track the scheme validation

MSLilah:
@jpinsonault @KevinCorcoran The most recent commit is a docs change only, so I'm going to go ahead and merge even though Travis isn't finished.

cprice404:
@camlow325 I would appreciate it if you would review the docs when you get a chance.  (I need to do so as well when I am back in the office next week).

documentation/status-proxy-service.md
```diff
@@ -0,0 +1,123 @@
...
+### Configuring the status-service for plaintext access
+
+If you would like to configure your status service to be on a plaintext port,
+your trapperkeeper configuration might look something like this:
+
+```
+webserver: {
+  default: {
+    ssl-port: 9001
+    ssl-cert: ./ssl/certs/localhost.pem
```
camlow325, line 46: I'd lean toward using fully-qualified paths in these examples.

documentation/status-proxy-service.md
```diff
@@ -0,0 +1,123 @@
...
+    default-server: true
+  }
+
+  status: {
+    port: 8080
+  }
+}
+
+web-router-service: {
+    "puppetlabs.trapperkeeper.services.status.status-service/status-service": {
```
camlow325, line 58: Cosmetic - would be good for the alignment in the `web-router-service` settings to be consistent with those in the `webserver` section - either 2 or 4 leading spaces.

documentation/status-proxy-service.md
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    "puppetlabs.trapperkeeper.services.status.status-proxy-service/status-proxy-service": {
+      route: /status-proxy
+      server: status-proxy
+    }
+}
+
+status-proxy: {
+  proxy-target-url: "https://localhost:9001/status"
```
camlow325, line 104: Can't remember the answer to this offhand, but it may be the case that the Jetty http proxy client would validate that the hostname used in the target URL matches the CN or a SubjectAltName in the server certificate - the server certificate presented by the Jetty server running at port 9001 in this example.  If that is the case, probably worth mentioning that the hostname in the `proxy-target-url` must match the CN or a SubjectAltName in the server certificate since "localhost" probably wouldn't be on the certificate used in production even though a host name of "localhost" would likely still be resolvable from the machine doing the proxying.  Maybe would be good to use something other than "localhost" throughout these examples to make that distinction more clear, e.g., "myhostname".

documentation/status-proxy-service.md
```diff
@@ -0,0 +1,123 @@
...
+The important things to note are:
+* The status proxy service and the status service are running on separate
+  webservers and ports
+* There is a new section in the config, `status-proxy`, with:
+  * a url pointing the proxy to the status service
+  * SSL information that matches the SSL information for the webserver that the
+    status service is running on
+
+With this configuration, requests that would normally be made to
+`https://localhost:9001/status` with an SSL cert can be made to
```
camlow325, line 122: Could say "can _also_ be made to..."

documentation/status-proxy-service.md
```diff
@@ -0,0 +1,123 @@
...
+    default-server: true
+  }
+
+  status-proxy: {
+    port: 8080
+  }
+}
+
+web-router-service: {
+    "puppetlabs.trapperkeeper.services.status.status-service/status-service": {
```
camlow325, line 92: I think you could also go with the slightly simpler syntax, which infers the "default" server:

```
  "puppetlabs.trapperkeeper.services.status.status-service/status-service": "/status"
```

documentation/status-proxy-service.md
```diff
@@ -0,0 +1,123 @@
...
+* The status proxy service and the status service are running on separate
+  webservers and ports
+* There is a new section in the config, `status-proxy`, with:
+  * a url pointing the proxy to the status service
+  * SSL information that matches the SSL information for the webserver that the
+    status service is running on
+
+With this configuration, requests that would normally be made to
+`https://localhost:9001/status` with an SSL cert can be made to
+`localhost:8080/status-proxy` over plaintext
```
camlow325, line 123: I think it would be clearer to use "_http://_localhost:8080/status-proxy" here.

documentation/status-proxy-service.md
```diff
@@ -0,0 +1,123 @@
...
+
+This config contains a new jetty server called `status` running on port 8080
+that is separate from the other, SSL enabled, server. In the 
+`web-router-service` section, the status service is configured to run on the
+`status` server, and be mounted at `/status`
+
+
+### Configuring the status-proxy-service
+
+If you would like to setup the status proxy service, your trapperkeeper
```
camlow325, line 73: Probably worth mentioning somewhere in here that the "status-proxy-service" would need to be added to the "bootstrap.cfg" as well.

camlow325:
@cprice404 @jpinsonault - I left a few comments on the docs.  Since this has been merged already, I'll leave it up to you to decide how you want to address these - maybe a MAINT follow-up PR?  I imagine @cprice404 will have some additional comments on these as well.

jpinsonault:
@camlow325 Thanks for the comments, I'll get a MAINT PR up once I've addressed your feedback

PR #23, opened 2015-06-17
(maint) Add dependency exclusion for :pedantic?

KevinCorcoran:
Starting a REPL in Cursive blows-up without this patch.


MSLilah:
:+1:

PR #24, opened 2015-06-18
(TK-231) Return 200 resp iff all services are up.

aperiodic:
Change the status service's handler to return a 200 OK HTTP response if
and only if all the services being queried have the :is-running field of
their status set to :true. If any service has any other value for this
field, or is missing the field from that response entirely, return a 503
Service Unavailable response.

### DIFF
--- a/test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
+++ b/test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
@@ -42,13 +42,19 @@
   [[:StatusService register-status]]
   (init [this context]
     (register-status "bar" "0.1.0" 1 (fn [level] {:status (str "bar status 1 " level)
-                                                  :is-running :false}))
+                                                  :is-running :true}))
     context))
 
 (defservice baz-service
   [[:StatusService register-status]]
   (init [this context]
     (register-status "baz" "0.2.0" 1 (fn [level] "baz"))
+    context))
+
+(defservice fail-service
+  [[:StatusService register-status]]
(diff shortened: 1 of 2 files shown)


test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -151,6 +157,23 @@
...
               (json/parse-string (slurp (:body resp)))))))))
 
+(deftest status-code-test
+  (testing "returns 503 response code when a service is not running"
+    (with-status-service app
+      [bar-service fail-service]
+      (let [{:keys [status]} (http-client/get "http://localhost:8180/status/v1/services/fail")]
+        (is (= 503 status)))
+      (let [{:keys [status]} (http-client/get "http://localhost:8180/status/v1/services")]
+        (is (= 503 status)))))
```
cprice404: Could we add a test here that validates we get a 200 for `/status/v1/services/bar`?  (Assuming we agree that's the desired behavior :) )

cprice404:
+1 other than maybe adding one more assertion to the test

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
MSLilah: :+1: on this suggestion

MSLilah:
I'm also :+1: barring the addition of another test assertion

aperiodic:
@cprice404 @fpringvaldsen so updated.

PR #25, opened 2015-06-19
(TK-231) Set JSON content type on responses

aperiodic:
Change the order of middleware application so that the
wrap-json-response middleware adds the content type header first,
resulting in the desired 'application/json' value.


rlinehan:
@aperiodic this has a bunch of what appear to be legit test failures.

aperiodic:
@rlinehan indeed. My apologies! I've updated the PR to a version that passes all tests for me.

PR #26, opened 2015-06-22
(TK-175) Add error handling for status check functions

vilmibm:
This commit adds special handling around the calling of status check
functions. Specifically, it handles these cases:
- Status check function times out
- Status check function throws an Exception
- Status check function returns something that doesn't conform to the
  callback response schema

In each error case, is_running is set to unknown and a descriptive
status key is set.

Note that none of these cases are reflected in the eventual TK status
service HTTP responses as that was not part of the specification of this
ticket.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/status/status_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_core.clj
@@ -203,28 +203,28 @@
     service :- [ServiceInfo]
     level :- ServiceStatusDetailLevel
     service-status-version :- (schema/maybe schema/Int)]
-  (let [status (if (nil? service-status-version)
-                   (last (sort-by :service-status-version service))
-                   (first (filter #(= (:service-status-version %)
-                                    service-status-version)
-                            service)))]
-      (when (nil? status)
-        (throw+ {:type    :service-status-version-not-found
-                 :message (str "No status function with version "
-                            service-status-version
-                            " found for service "
-                            service-name)}))
-      (let [timeout (check-timeout level)
-            callback-resp (guarded-status-fn-call (:status-fn status) level timeout)
-            data (:status callback-resp)
-            state (if-not (schema/check State (:state callback-resp))
-                         (:state callback-resp)
-                         :unknown)]
-        {:service_version (:service-version status)
-         :service_status_version (:service-status-version status)
-         :detail_level level
-         :state state
-         :status data}))))
+   (let [status (if (nil? service-status-version)
+                  (last (sort-by :service-status-version service))
+                  (first (filter #(= (:service-status-version %)
+                                     service-status-version)
+                                 service)))]
+     (when (nil? status)
+       (throw+ {:type    :service-status-version-not-found
+                :message (str "No status function with version "
+                              service-status-version
+                              " found for service "
+                              service-name)}))
+     (let [timeout (check-timeout level)
+           callback-resp (guarded-status-fn-call (:status-fn status) level timeout)
+           data (:status callback-resp)
+           state (if-not (schema/check State (:state callback-resp))
+                   (:state callback-resp)
+                   :unknown)]
+       {:service_version (:service-version status)
+        :service_status_version (:service-status-version status)
+        :detail_level level
+        :state state
+        :status data}))))
 
 (schema/defn ^:always-validate call-status-fns :- ServicesStatus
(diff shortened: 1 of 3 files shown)


aperiodic:
@nathanielksmith needs a rebase.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -127,6 +139,47 @@
...
+    (validation-error-explain check-result)
+    check-result))
+
+(schema/defn ^:always-validate guarded-status-fn-call
+  "Given a status check function, a status detail level, and a timeout in
+  seconds, this function calls the status function and handles three types of
+  errors:
+
+  * Status check timed out
+  * Status check through an Exception
```
aperiodic: Homonym mismatch.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -127,6 +139,47 @@
...
+  string describing the error."
+  [status-fn :- StatusFn
+   level :- ServiceStatusDetailLevel
+   timeout :- schema/Int] :- StatusCallbackResponse
+   (let [unknown-response (fn [status] {:is-running :unknown
+                                        :status status})
+         timeout-response (unknown-response (format "Status check timed out after %s seconds" timeout))]
+     (with-timeout timeout timeout-response
+       (try
+         (let [result (status-fn level)]
```
aperiodic: `status` seems like a better name then the very generic label of `result`.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -127,6 +139,47 @@
...
+           (if-let [check-result (schema/check StatusCallbackResponse result)]
+             (unknown-response (format "Status check malformed: %s" (maybe-explain check-result)))
```
aperiodic: I think `check-result` is also a mechanical description that doesn't give the reader any more information. What about `schema-failures` or `schema-problems`?

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -142,7 +195,7 @@
     service :- [ServiceInfo]
     level :- ServiceStatusDetailLevel
     service-status-version :- (schema/maybe schema/Int)]
-    (let [status (if (nil? service-status-version)
+   (let [status (if (nil? service-status-version)
```
aperiodic, line 206: If the let is moving then the entire binding vector & body need to shift as well, to maintain indentation. That should probably be done in a separate commit.

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -51,6 +51,21 @@
...
+
+(defservice broken-service
+  [[:StatusService register-status]]
+  (init [this context]
+        (register-status "broken" "0.1.0" 1 (fn [level] (throw (Exception. "don't"))))))
+
+(defservice punk-service
+  [[:StatusService register-status]]
+  (init [this context]
+        (register-status "punk" "0.1.0" 1 (fn [level] {:non :conforming}))))
```
aperiodic: Isn't this functionally the same as the `baz-service` above?

vilmibm:
@aperiodic updated.

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -178,6 +188,28 @@
...
+        (is (re-find #"timed out" (get body "status"))))))
+
+  (testing "handles case when a status check throws an exception"
+    (let [resp (http-client/get "http://localhost:8180/status/v1/services/broken?level=critical")
+            body (json/parse-string (slurp (:body resp)))]
+        (is (= 503 (:status resp)))
+        (is (= "unknown"
+               (get body "state")))
+        (is (re-find #"exception.*don't" (get body "status")))))))
+
```
aperiodic, line 220: I think there should still be an explicit test of the behavior when a service returns a non-conformant status response, it just doesn't need a new service to do it.

aperiodic:
@nathanielksmith I'd like to see the third test case from the status-check-error-handling-test added back in, but the rest of this looks good. I'll merge once that case is back in.

vilmibm:
@aperiodic done

PR #27, opened 2015-06-22
(TK-229) Rename is_running to state

pcarlisle:
Rename `is_running` to `state` and possible values from `true`, `false`, and `unknown`
to `running`, `error`, and `unknown`.


aperiodic:
LGTM

PR #28, opened 2015-06-24
(maint) Fix dependency issue

rlinehan:
Other projects using trapperkeeper-status were exploding because it was
transitively pulling in a really old version of the java servlet. This commit
excludes that dependency, so this should no longer be an issue.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -15,7 +15,7 @@
                  [cheshire "5.3.1"]
                  [prismatic/schema "0.4.0"]
                  [ring/ring-json "0.3.1" :exclusions [ring/ring-core]]
-                 [ring/ring-defaults "0.1.5"]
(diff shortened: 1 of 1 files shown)


juxtin:
:+1: for getting rid of those explosions

aperiodic:
@rlinehan this appears to introduce some test failures.

aperiodic:
Although, I can't reproduce those locally. @holguinj, could you try this branch out and see if it works for you?

juxtin:
Tests pass on my machine

rlinehan:
@holguinj @aperiodic I think this is because there were some other conflicts in another PR that got merged and were causing the tests to fail. Hopefully if I just rebase and force push it should be fine? We'll see.

PR #30, opened 2015-06-26
(TK-245) Schema validation for status-proxy-service

jpinsonault:


test/puppetlabs/trapperkeeper/services/status/status_proxy_service_test.clj
```diff
@@ -30,7 +30,9 @@
                         :host "0.0.0.0"}
    :status-proxy       {:proxy-target-url "https://0.0.0.0:9001/ssl-status"
                         :ssl-opts common-ssl-config}
-   :web-router-service {:puppetlabs.trapperkeeper.services.status.status-proxy-service/status-proxy-service "/status-proxy"}})
+   :web-router-service {:puppetlabs.trapperkeeper.services.status.status-proxy-service/status-proxy-service "/status-proxy"}
+   ; Tell TK to not log errors to the screen. For tests with intentionally broken services
```
jpinsonault: the `invalid-proxy-config-throws` test outputs TK error log messages to the screen when running `lein test` for the intentionally improperly configured services. I decided to mute them with a `logback.xml` file. Not sure if this is reasonable.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -52,6 +52,12 @@
 (def SemVerVersion
   (schema/pred semver/valid-format? "semver"))
 
+(def StatusProxyConfig
+  {:proxy-target-url schema/Str
```
MSLilah, line 57: You can use `schema/pred` to use your `validate-protocol` function to validate the `proxy-target-url` here in the `StatusProxyConfig` schema itself, rather than separately calling your `validate-protocol` function. I'm not sure how nicely that would play with the exception you're throwing in `validate-protocol` though.

src/puppetlabs/trapperkeeper/services/status/status_proxy_service.clj
```diff
@@ -8,15 +11,20 @@
...
-          path (.getPath target-url)]
-      (add-proxy-route
-        this
-        {:host host
-         :port port
-         :path path}
-        {:ssl-config (get-in-config [:status-proxy :ssl-opts])
-         :scheme     :https}))
+    (let [status-proxy-config (get-in-config [:status-proxy])]
+      (schema/validate StatusProxyConfig status-proxy-config)
```
MSLilah: I'm wondering if we should extract most of this work into a function in the core namespace? I know we've been trying to keep service function definitions as small as possible since they're really hard to debug because macros, and it seems like we're doing enough work here that it would be worth extracting.

test/puppetlabs/trapperkeeper/services/status/status_proxy_service_test.clj
```diff
@@ -20,9 +20,9 @@
    :ssl-ca-cert (str dev-resources "/ssl/certs/ca.pem")})
 
 (def ssl-status-service-config
-  {:webserver (merge common-ssl-config
-                {:ssl-host "0.0.0.0"
-                 :ssl-port 9001})
+  {:webserver          (merge common-ssl-config
```
MSLilah: I don't think this indentation conforms to our indentation standards, I don't think we should be aligning the values in a map with each other.

test/puppetlabs/trapperkeeper/services/status/status_proxy_service_test.clj
MSLilah: @jpinsonault I think the way we would normally mute these log messages is by wrapping the relevant portion of the test in a call to `with-test-logging`. This captures log messages from TK and allows you to write assertions that certain messages were logged, and it also has the effect of stopping those messages from being logged to the console.

test/puppetlabs/trapperkeeper/services/status/status_proxy_service_test.clj
```diff
@@ -161,3 +163,47 @@
             (doall (map #(http-client/get % common-ssl-config) good-non-status-endpoint-requests))
             (is (= (count good-non-status-endpoint-requests)
                   (deref non-status-endpoint-counter)))))))))
+
+(deftest invalid-proxy-config-throws
+  (testing "Missing proxy-target-url throws schema error"
+    (is (thrown-with-msg?
+          clojure.lang.ExceptionInfo
+          #"Value does not match schema.*:proxy-target-url"
+          (let [bad-config (update-in
```
MSLilah: Nitpick, but I'd prefer if the generation of the bad config here and below was done before the test assertion, since it's the result of using that bad config that's being tested rather than the generation of that config.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
jpinsonault, line 57: I decided to go with the stand alone validation function, because when I used `schema/pred` the error it gave wasn't very clear about what the actual problem was. Maybe there's a way to put a custom error message when using a predicate though?

src/puppetlabs/trapperkeeper/services/status/status_proxy_service.clj
jpinsonault: Yeah, sounds like a good idea

test/puppetlabs/trapperkeeper/services/status/status_proxy_service_test.clj
jpinsonault: Good point. I'll track down the intellij setting that forces it =)

test/puppetlabs/trapperkeeper/services/status/status_proxy_service_test.clj
jpinsonault: Good call. The less irrelevant stuff in the assert the better.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
MSLilah, line 57: I don't think there is. I'm fine with leaving this how it is now, then, in the interest of giving the user a useful error message.

jpinsonault:
@fpringvaldsen Addressed your feedback:
- Switched to `with-test-logging`
- Extracted the core bits from the `init` function into a function in the `core` ns
- Fixed indentation issue
- Moved the `bad-config` generation out of the asserts

MSLilah:
@jpinsonault Looking good! I had one minor bit of feedback on validating the new function you added. That should be a quick change, and aside from that I'm :+1:!

jpinsonault:
@fpringvaldsen Fixed up the schema validation

MSLilah:
:shipit:

src/puppetlabs/trapperkeeper/services/status/status_proxy_service.clj
```diff
@@ -1,22 +1,15 @@
...
-      (add-proxy-route
-        this
-        {:host host
-         :port port
-         :path path}
-        {:ssl-config (get-in-config [:status-proxy :ssl-opts])
-         :scheme     :https}))
+    (let [status-proxy-config (get-in-config [:status-proxy])
+          {proxy-target :proxy-target
+           proxy-options :proxy-options} (get-proxy-route-info status-proxy-config)]
```
KevinCorcoran: Wow, that is some Clojure voodoo I did not know existed.

You can use the `:keys` destructuring magic here, it's a bit more concise and common:

``` clj
    (let ...
         {:keys [proxy-target proxy-options]} (get-proxy-route-info status-proxy-config)]
```

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -98,6 +105,16 @@
...
 
+(defn validate-protocol
+  "Throws if the protocol is not http or https"
+  [url]
+  (let [protocol (.getProtocol url)]
+    (if-not (contains? #{"http" "https"} protocol)
+      (throw (IllegalStateException.
+               (format (str "The proxy-target-url '%s' has an unsupported "
+               "protocol '%s'. Must be either http or https")
+                 (str url) protocol))))))
```
KevinCorcoran: indentation is off on the two lines above

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -98,6 +105,16 @@
     (when (or differing-svc-version? differing-status-version?)
       (throw (IllegalStateException. error-message)))))
 
+(defn validate-protocol
```
KevinCorcoran: We typically name functions like this with a `!` at the end to indicate that they don't return a meaningful value and are only called for their side effects - i.e. `validate-protocol!`.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -98,6 +105,16 @@
...
+  "Throws if the protocol is not http or https"
+  [url]
+  (let [protocol (.getProtocol url)]
+    (if-not (contains? #{"http" "https"} protocol)
+      (throw (IllegalStateException.
```
KevinCorcoran: `IllegalArgumentException` seems more appropriate than `IllegalStateException` in this case.

src/puppetlabs/trapperkeeper/services/status/status_proxy_service.clj
jpinsonault: Ditto! Voodoo and magic

KevinCorcoran:
:chart_with_upwards_trend: looks good except for nitpicks

src/puppetlabs/trapperkeeper/services/status/status_core.clj
jpinsonault: Makes sense, thanks

jpinsonault:
@KevinCorcoran Adressed your feedback

PR #31, opened 2015-06-27
 (TK-230) Implement "simple" endpoint for plaintext responses

vilmibm:
This commit adds two new endpoints to tk-status that support no query
parameters and return only simple plaintext messages and relevant status
codes.

Note that in order to handle the potential for either plaintext or
json responses in our error handling middleware, there is a mix of
compojure and comidi functions when definiing tk-status's ring handler.
We use compojure's functions to compose handlers which have been
generated by comidi and then had middleware applied to them.

We required this because we could no longer apply the same list of
middleware to the entire ring handler (since now the json routes and
plaintext routes required different sets of middleware).

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/status/ringutils.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/ringutils.clj
@@ -2,47 +2,85 @@
 ;; probably be moved into a shared library or some such.
 (ns puppetlabs.trapperkeeper.services.status.ringutils
   (:require [clojure.tools.logging :as log]
+            [cheshire.core :as json]
+            [ring.util.response :as response]
+            [schema.core :as schema]
             [slingshot.slingshot :refer [try+]]
             [puppetlabs.kitchensink.core :as ks]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-;;; Public
+;;; Schemas
 
-(defn wrap-request-data-errors
+(def ResponseType (schema/enum :json :plain))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Public
+(defn json-response [status body]
+  (-> body
+      json/encode
+      response/response
+      (response/status status)
+      (response/content-type "application/json; charset=utf-8")))
+
+(defn plain-response [status body]
+  (-> body
+      response/response
+      (response/status status)
+      (response/content-type "text/plain; charset=utf-8")))
+
+(schema/defn ^:always-validate wrap-request-data-errors
   "A ring middleware that catches slingshot errors of :type
   :request-dava-invalid and returns a 400."
-  [handler]
-  (fn [request]
-    (try+ (handler request)
-          (catch #(contains? #{:request-data-invalid :service-status-version-not-found}
-                             (:type %)) e
-            {:status 400
-             :body e}))))
+  [handler type :- ResponseType]
+  (let [code 400
+        response (fn [e]
+                   (case type
+                     :json (json-response code e)
+                     :plain (plain-response code (:message e))))]
+    (fn [request]
+      (try+ (handler request)
+            (catch
+                #(contains? #{:request-data-invalid :service-status-version-not-found}
+                            (:type %))
+                e
+              (response e))))))
 
-(defn wrap-schema-errors
+(schema/defn ^:always-validate wrap-schema-errors
   "A ring middleware that catches schema errors and returns a 500
   response with the details"
-  [handler]
-  (fn [request]
-    (try (handler request)
-         (catch clojure.lang.ExceptionInfo e
-           (let [message (.getMessage e)]
-             (if (re-find #"does not match schema" message)
-               {:status 500
-                :body {:type :application-error
-                       :message (str "Something unexpected happened: "
-                                     (select-keys (.getData e) [:error :value :type]))}}
-               ;; re-throw exceptions that aren't schema errors
-               (throw e)))))))
+  [handler type :- ResponseType]
+  (let [code 500
+        response (fn [e]
+                   (let [msg (str "Something unexpected happened: "
+                                  (select-keys (.getData e) [:error :value :type]))]
+                     (case type
+                       :json (json-response code
+                               {:type :application-error
+                                :message msg})
+                       :plain (plain-response code msg))))]
+    (fn [request]
+      (try (handler request)
+           (catch clojure.lang.ExceptionInfo e
+             (let [message (.getMessage e)]
+               (if (re-find #"does not match schema" message)
+                 (response e)
+                 ;; re-throw exceptions that aren't schema errors
+                 (throw e))))))))
 
-(defn wrap-errors
+(schema/defn ^:always-validate wrap-errors
   "A ring middleware that catches all otherwise uncaught errors and
   returns a 500 response with the error message"
-  [handler]
-  (fn [request]
-    (try (handler request)
-         (catch Exception e
-           (log/error e "Error on server")
-           {:status 500
-            :body {:type :application-error
-                   :message (str "Error on server: " e)}}))))
+  [handler type :- ResponseType]
+  (let [code 500
+        response (fn [e]
+                   (let [msg (str "Error on server: " e)]
+                     (case type
+                       :json (json-response code
+                               {:type :application-error
+                                :message msg})
+                       :plain (plain-response code msg))))]
+    (fn [request]
+      (try (handler request)
+           (catch Exception e
+             (log/error e "Error on server")
+             (response e))))))
--- a/src/puppetlabs/trapperkeeper/services/status/status_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_core.clj
@@ -1,9 +1,9 @@
 (ns puppetlabs.trapperkeeper.services.status.status-core
   (:require [schema.core :as schema]
             [schema.utils :refer [validation-error-explain]]
-            [ring.middleware.json :as ring-json]
             [ring.middleware.defaults :as ring-defaults]
             [slingshot.slingshot :refer [throw+]]
+            [compojure.core :as compojure]
             [puppetlabs.comidi :as comidi]
             [puppetlabs.kitchensink.core :as ks]
             [puppetlabs.trapperkeeper.services.status.ringutils :as ringutils]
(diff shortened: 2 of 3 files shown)


src/puppetlabs/trapperkeeper/services/status/ringutils.clj
```diff
@@ -1,48 +1,85 @@
 ;; TODO: Much of this code is copied from other projects. It should
 ;; probably be moved into a shared library or some such.
 (ns puppetlabs.trapperkeeper.services.status.ringutils
-  (:require [clojure.tools.logging :as log]
+  (:require [schema.core :as schema]
+            [clojure.tools.logging :as log]
+            [cheshire.core :as json]
+            [ring.util.response :as response]
```
aperiodic, line 6: what's the ordering here?

vilmibm, line 6: none. do we have a standard?

src/puppetlabs/trapperkeeper/services/status/ringutils.clj
```diff
@@ -1,48 +1,85 @@
...
+       json/encode
+       response/response
+       (response/status ~status)
+       (response/content-type "application/json; charset=utf-8")))
+
+(defmacro with-plain-response [status & body]
+  `(-> (do ~@body)
+       response/response
+       (response/status ~status)
+       (response/content-type "text/plain; charset=utf-8")))
```
aperiodic: I don't understand why these are macros instead of just functions called `json-response` and `plain-response` that take a status and a message body.

src/puppetlabs/trapperkeeper/services/status/ringutils.clj
vilmibm, line 6: I updated it to be more sane (clojure, then alphabet, then puppetlabs)

src/puppetlabs/trapperkeeper/services/status/ringutils.clj
vilmibm: they're from another library where they actually wrap bodies. I didn't go back and make them functions. I can do that.

src/puppetlabs/trapperkeeper/services/status/ringutils.clj
```diff
@@ -2,47 +2,84 @@
...
+  [handler type :- ResponseType]
+  (let [code 400
+        response (fn [e]
+                   (case type
+                     :json (with-json-response code e)
+                     :plain (with-plain-response code (:message e))))]
+    (fn [request]
+      (try+ (handler request)
+            (catch #(contains? #{:request-data-invalid :service-status-version-not-found}
+                               (:type %)) e
```
aperiodic: The binding of the caught exception to `e` is pretty easy to miss here. I think it's always bad style to have a new argument on the same line as the ending of a multi-line argument form; if you've got a multi-line argument then it should be the only thing on its lines.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -55,6 +55,21 @@
...
+     result#))
+
+(defn- maybe-explain
+  "Given the result of a call to schema.core/check, potentially unwrap it with
+  validation-error-explain if it is a ValidationError object. Otherwise, pass
+  the argument through."
+  [schema-failure]
+  (if (instance? schema.utils.ValidationError schema-failure)
+    (validation-error-explain schema-failure)
+    schema-failure))
```
aperiodic, line 71: Haven't these both been defined a few places now? We should put them in KS so we can eventually get rid of them here (obv. don't block this PR on that, though).

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -87,11 +102,11 @@
 ;;; Public
 
 (schema/defn ^:always-validate nominal?
-  [status :- ServiceStatus]
+  [status :- ServiceStatus] :- schema/Bool
   (= (:state status) :running))
 
 (schema/defn ^:always-validate all-nominal?
-  [statuses :- ServicesStatus]
+  [statuses :- ServicesStatus] :- schema/Bool
```
aperiodic: Schema annotations for the return value of a function go after the function's name, not its argument vector.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
aperiodic, line 71: Oh, I see, maybe-explain got moved up. `with-timeout` is used in another project, though (jdbc-util), so that still seems like a good candidate for KS.

vilmibm, line 71: This is just a move.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
vilmibm: this explains a lot.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -264,20 +264,64 @@
...
+   * if all of the statuses have the same :state, returns that :state
+   * if not all of the statuses are the same, returns :unknown"
+  [statuses :- ServicesStatus] :- State
+  (let [state-set (->> statuses
+                       vals
+                       (map :state)
+                       set)]
+    (if (= 1 (count state-set))
+      (first state-set)
+      :unknown)))
```
aperiodic: I don't think this is the right behavior. If some of the services are in an :error state, but some are :running, I think the summary should be that the whole system is in an :error state. If none of them are in :error, but at least one is :unknown, then I think it makes sense to report :unknown. So basically, the ordering is :error > :unknown > :running, and the summary behavior is to choose the greatest from among the reported states by that ordering.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -264,20 +264,64 @@
...
+  "Given a service status, returns an appropriate HTTP status code"
+  [status :- ServiceStatus] :- schema/Int
+  (if (nominal? status) 200 503))
+
+(schema/defn ^:always-validate statuses->code
+  "Given a map of service statuses, returns an appropriate HTTP status code."
+  [statuses :- ServicesStatus] :- schema/Int
+  (if (all-nominal? statuses)
+    200
+    503))
```
aperiodic: Why three lines for the if here but all on one line above? The extra `all-` in the second example doesn't put it anywhere close to the line limit.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -264,20 +264,64 @@
...
+  [path status-fns]
+  (comidi/context path
+     (comidi/GET "" _
+       (let [statuses (call-status-fns status-fns :critical)]
+         (ringutils/with-plain-response (statuses->code statuses)
+           (-> statuses
+               summarize-states
+               name))))
+
+     (comidi/GET ["/" :service-name] [service-name :as _]
```
aperiodic: If this is functionally identical to compojure you don't have to bind the whole request w/ `:as _`; you can just leave that part out if you're not using it.

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -244,6 +250,65 @@
...
+        (is (= 200 (:status resp)))
+        (is (= "running" (slurp (:body resp))))))
+    (testing "ignores alphabetic service_status_version"
+      (let [resp (http-client/get (str "http://localhost:8180/status/v1/"
+                                    "simple/foo?service_status_version=abc"))]
+        (is (= 200 (:status resp)))
+        (is (= "running" (slurp (:body resp))))))
+    (testing "ignores non-existent service_status_version"
+      (let [resp (http-client/get (str "http://localhost:8180/status/v1/"
+                                    "simple/foo?service_status_version=3"))]
```
aperiodic: This indentation style isn't consistent with the rest of your PR (and also above on line 263).

src/puppetlabs/trapperkeeper/services/status/status_core.clj
vilmibm: It's definitely required in the GET "" case; I get an unsupported binding exception without _. The latter one, yes, I can leave :as out

aperiodic: Yeah, there always needs to be some kind of binding form, even if it's just `_` or an empty vector.

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -244,6 +250,65 @@
...
+            (is (= 200 (:status resp)))
+            (is (= "running" (slurp (:body resp))))))
+        (testing "and it is not :running"
+          (let [resp (http-client/get "http://localhost:8180/status/v1/simple/baz")]
+            (is (= 503 (:status resp)))
+            (is (= "unknown" (slurp (:body resp))))))
+        (testing "and it does not exist"
+          (let [resp (http-client/get "http://localhost:8180/status/v1/simple/kafka")]
+            (is (= 404 (:status resp)))
+            (is (= "not found: kafka" (slurp (:body resp))))))))))
```
aperiodic, line 308: To round this out it seems like we should have a test for when the single service reports an explicit :error state.

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -244,6 +244,65 @@
...
+          (let [resp (http-client/get "http://localhost:8180/status/v1/simple")]
+            (is (= 503 (:status resp)))
+            (is (= "error" (slurp (:body resp)))))))
+
+      (testing "and a service is :unknown"
+        (with-status-service app
+          [foo-service baz-service]
+          (let [resp (http-client/get "http://localhost:8180/status/v1/simple")]
+            (is (= 503 (:status resp)))
+            (is (= "unknown" (slurp (:body resp))))))))
```
aperiodic, line 288: These tests seem a little sparse given that there's no distinction between the current behavior and what I think the behavior should be. Probably just a test of what happens when there's a :running service, an :error service, and a service that doesn't report its status correctly is enough to cover the gap.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -264,20 +264,64 @@
...
+  [statuses :- ServicesStatus]
+  (let [state-set (->> statuses
+                       vals
+                       (map :state)
+                       set)]
+    (cond
+      (state-set :error) :error
+      (state-set :unknown) :unknown
+      (state-set :running) :running
+      :else :runing)))
```
aperiodic: So first off there's a typo here, but it seems safest to report :unknown if we get here (given that we don't think it should even be possible).

aperiodic:
@nathanielksmith I'm finished reviewing; once the outstanding comments are addressed, I think it's good to merge.

vilmibm:
@aperiodic all updates done.

aperiodic:
LGTM
thumbs-up

PR #34, opened 2015-07-17
(TK-247) Path traversal attack tests

jpinsonault:
See https://www.owasp.org/index.php/Path_Traversal for details on the type of vulnerability
I initially thought this was going to be something somewhat complicated, but I think that since the status service doesn't have any url parameters relating to file paths, the only concern is that someone might try to use the proxy and a (non-existent) loophole in jetty to access something like `http://localhost:9000/status-proxy/v1/../../something-secret/` and get access to `https://localhost:9001/something-secret/`

### DIFF
--- a/test/puppetlabs/trapperkeeper/services/status/status_proxy_service_test.clj
+++ b/test/puppetlabs/trapperkeeper/services/status/status_proxy_service_test.clj
@@ -118,7 +118,9 @@
    :body   (str (swap! counter inc))})
 
 (deftest proxy-only-proxies-what-it-should-test
-  (testing "status-proxy-service doesn't proxy things it shouldn't"
+  (testing "status-proxy-service doesn't proxy things it shouldn't.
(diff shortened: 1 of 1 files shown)


MSLilah:
It seems to me that these tests are testing the underlying functionality of Jetty and our `add-proxy-route` function rather than anything specific to the status service, so I think it might be better for these tests to live in the TK-Jetty9 repo rather than in the trapperkeeper-status repo.

camlow325:
The tests seem reasonable to me, but I'd agree with @fpringvaldsen that it would seem better to have them live in tk-jetty9.  Either way, I guess we're ultimately testing Jetty and not any PL-specific functionality but it also seems reasonable to create/maintain these tests because of the potential security risks if this were to break.

jpinsonault:
Ok, that makes sense. I'll move them to the tk-jetty9 repo. Thanks guys!

PR #35, opened 2015-07-20
(maint) Log Exceptions thrown by callbacks

KevinCorcoran:


MSLilah:
I think it would be worth adding a test to ensure an exception is logged. Other than that I'm :+1:

KevinCorcoran:
In general, I'm opposed to writing tests for logging.  In my experience, such tests almost always end up adding little or no value in the long run, and therefore don't seem worth the effort.  Logging seems like one of those things that's better validated manually (i.e. looking at the output).

Perhaps we could get another opinion or two on this?  I'm fine with writing a test if there's agreement that it's worthwhile.

rlinehan:
I'm :+1: and could go either way on a test.

I've seen a number of instances where a change accidentally removed logging, and it wasn't noticed because no one was testing that error case aside from the tests which used `with-test-logging`. I think when the only test for something is testing log output, then that's pretty terrible, I'm less convinced that it's terrible when it's in addition to other testing, but I guess I feel generally "meh" about it. I agree with Kevin that for the effort involved, the tests aren't usually very good and don't provide a ton of worth aside from testing that the log message matches, which often has to be changed at some point.

That probably wasn't super helpful; I don't have a super strong opinion, sorry.

KevinCorcoran:
@rlinehan @fpringvaldsen Thanks for the feedback.  I added a commit which adds a  couple assertions to the existing tests which already used `with-test-logging` to check for the error message.  Seemed like a reasonable middle ground.

KevinCorcoran:
Had to trim it down to one assertion.  Apparently, in the test case which times out, an `InterruptedException` doesn't always get thrown (which is why the added assertion about an error being logged was sometimes failing), even though `call-status-fn-for-service` returns the correct value indicating a timeout either way.  I don't fully understand how that works, but I decided to cut my losses and not dig any deeper.

MSLilah:
:+1:

rlinehan:
:+1:

PR #37, opened 2015-11-04
(TK-290) add starting and stopping states

wkalt:
This adds starting and stopping to the set of possible states for the status
service.

### DIFF
--- a/test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
+++ b/test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
@@ -58,6 +58,18 @@
     (register-status "fail" "4.2.0" 1 (fn [level] {:status "wheee", :state :error}))
     context))
 
+(defservice starting-service
+  [[:StatusService register-status]]
+  (init [this context]
+        (register-status "starting" "6.6.6" 1 (fn [level] {:status "foo"
+                                                           :state :starting}))))
+
+(defservice stopping-service
+  [[:StatusService register-status]]
+  (init [this context]
+        (register-status "stopping" "6.6.6" 1 (fn [level] {:status "bar"
+                                                           :state :stopping}))))
+
 (defservice slow-service
   [[:StatusService register-status]]
   (init [this context]
@@ -282,6 +294,13 @@
             (is (= 503 (:status resp)))
             (is (= "error" (slurp (:body resp)))))))
 
+      (testing "and one service is :starting while another is :stopping"
+        (with-status-service app
+          [foo-service bar-service starting-service stopping-service]
+          (let [resp (http-client/get "http://localhost:8180/status/v1/simple")]
+            (is (= 503 (:status resp)))
+            (is (= "stopping" (slurp (:body resp)))))))
+
       (testing "and a service is :unknown"
         (with-status-service app
           [foo-service baz-service]
@@ -291,7 +310,7 @@
 
     (testing "for a single service"
       (with-status-service app
-        [foo-service baz-service fail-service]
+        [foo-service baz-service fail-service starting-service]
         (testing "and it is :running"
           (let [resp (http-client/get "http://localhost:8180/status/v1/simple/foo")]
             (is (= 200 (:status resp)))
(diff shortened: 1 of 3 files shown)


test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -280,7 +286,14 @@
           [foo-service baz-service fail-service]
           (let [resp (http-client/get "http://localhost:8180/status/v1/simple")]
             (is (= 503 (:status resp)))
-            (is (= "error" (slurp (:body resp)))))))
+            (is (= "unknown" (slurp (:body resp)))))))
```
wkalt: @nathanielksmith note this test change: previously if any service reported an error, /simple would return :error. Based on our discussion in hipchat I've changed it to return :unknown in that case. The behavior now agrees with the summarize-states docstring, but we should be sure it's what we want.

cprice404:
@pcarlisle do you think your team could handle review/merge/release on this?  I'm +1 on the spirit of the changes as described on the mailing list.

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
aperiodic: I don't agree with this change. Seems like we're trying to soft-pedal things in a way that'll be confusing for users. What's the argument for reporting a status that potentially none of the services reported? Why do we want to collapse so many different status combinations into just returning "unknown"? Why would users rather see "unknown" and have to investigate rather than seeing the most severe status from among the reported statuses?

vilmibm: pinging @aperiodic since i think he might remember better than me the definitive thing

aperiodic: @nathanielksmith what definitive thing? Did you mean to ping @wkalt ?

vilmibm: @aperiodic The docs, a docstring, and the code all say something different about how this is supposed to work; i was just wondering if you remembered which was the "right" thing. i assume the code as it stands should be taken as definitive and the docstring/docs updated.

aperiodic: Haha whoops. Let me look at the other ones.

aperiodic: Hmm. I see how this happened now. The docstring does specify that `unknown` should be returned whenever the states don't all agree. The wire format documentation doesn't exactly specify the behavior -- it says `unknown` is used when there's a problem getting the status for some service, but doesn't say any more than that.

I think that the existing code captures the correct behavior, so the docstring should be updated. I think it's most useful to surface the "most severe" state that we got: if any one state is `error`, then that's what we should return; if none are `error`, then if any are `unknown` that gets returned; finally if all states are `running` then that's what we return. Equivalently, the states have a severity ranking and we return the highest-severity state that we got according to that ranking.

I'm not sure where `starting` and `stopping` fit into that ranking, but my initial impulse is that the ordering should be (from most to least severe):
- error
- unknown
- stopping
- starting
- running

@wkalt I am sorry for piling on all the questions under the assumption you wanted to change this behavior. I see now you were trying to match what you believed the correct behavior to be. I should have been less presumptive.

wkalt: @aperiodic no need to apologize, and my mistake for letting your comment slip by me somehow. If I understand your proposal, I think my only concern is that the precedence between starting and stopping becomes confusing when someone hits the /simple endpoint while one service is starting and another is stopping. That state really does feel more like "unknown" to me. I do agree that my implementation here should be changed back to reflect :error if any of the services return that state. Any additional thoughts on starting/stopping?

wkalt:
chatted with @aperiodic about this offline. I've changed the implementation of the /simple endpoint back to returning the first represented state among the services in the list :error, :unknown, :stopping, :starting, :running.

aperiodic:
LGTM :+1:

PR #38, opened 2015-12-11
Expose service method for retrieving registered status functions

adrienthebo:
The status/availability of one trapperkeeper service may depend on the presence of other services; this pull request adds a `get-status` method to the `StatusService` protocol (and implementing service) to provide a consistent interface for interacting with status functions across services.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/status/status_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_core.clj
@@ -211,6 +211,37 @@
              (log/error e error-msg)
              (unknown-response (format "%s: %s" error-msg e))))))))
 
+(schema/defn ^:always-validate matching-service-info :- ServiceInfo
+  "Find a service info entry matching the service-status-version. If
+  service-status-version is nil the most recent service info is returned."
+  [service-name :- schema/Str
+   service :- [ServiceInfo]
+   service-status-version :- (schema/maybe schema/Int)]
+   (let [status (if (nil? service-status-version)
+                  (last (sort-by :service-status-version service))
+                  (first (filter #(= (:service-status-version %)
+                                     service-status-version)
+                                 service)))]
+     (if (nil? status)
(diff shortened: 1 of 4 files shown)


cprice404:
:see_no_evil:

adrienthebo:
Added a couple more tests to make sure that different versions/levels can be used.

adrienthebo:
@cprice404 @holguinj I have other work in the current feature that depends on this feature; is there anything I can do to expedite getting this merged/released?

cprice404:
@holguinj @pcarlisle can someone from one of you guys' teams give this a review?  If you are OK with the API changes then I'm fine with it being merged / released.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -211,6 +211,37 @@
              (log/error e error-msg)
              (unknown-response (format "%s: %s" error-msg e))))))))
 
+(schema/defn ^:always-validate matching-service-info :- ServiceInfo
+  "Find a service info entry matching the service-status-version. If
+  service-status-version is nil the most recent service info is returned."
+  [service-name :- schema/Str
+   service :- [ServiceInfo]
+   service-status-version :- (schema/maybe schema/Int)]
```
juxtin, line 219: Unless there's a compelling reason not to, I think it's better to have two different arities of this function rather than take a final parameter that can be nil. Same with `get-status-fn`

juxtin, line 219: yup, I found the compelling reason. Disregard.

adrienthebo:
@holguinj was the nil/multiple arity question the main question you had about this PR, or are there other things that I need to address?

juxtin:
Still reviewing

juxtin:
Thanks, @adrienthebo!

adrienthebo:
<3

PR #42, opened 2016-02-04
(MAINT) add exclusion for obsolete servlet-api

cprice404:
ring-defaults has a dependency on an old version of the javax
servlet API.  This artifact has a different name than new versions
of the servlet API, which means that lein won't detect a conflict,
and you will end up with a classpath that has multiple copies of
the servlet API in it, which will cause all kinds of indecipherable
errors, and that ain't no good time for nobody.

This commit adds an exclusion to prevent that terrible old dependency
from being brought in transitively by consumers of tk-status.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 0.3.1
+
+This is a maintenance / bugfix release.
+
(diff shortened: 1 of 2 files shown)


nwolfe:
:metal:

PR #46, opened 2016-03-03
(TK-335) Make service lifecycle idempotent

vilmibm:
This commit allows the status service to be stopped and started (distinct from tk's `restart`). Prior to this
commit, the status service's `start` function would attempt to re-add
the status service's own function to `:status-fns`, resulting in a
validation error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 0.3.3
+
+This is a maintenance / bugfix release.
+
+* Allow trapperkeeper's `stop` and `start` functions to be called on the status
+  service without error by cleaning up context state manually.
+
 ## 0.3.2
 
 This is a maintenance / bugfix release.
--- a/src/puppetlabs/trapperkeeper/services/status/status_service.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_service.clj
@@ -23,7 +23,7 @@
     (assoc context :status-fns (atom {})))
 
   (start [this context]
-    (register-status this "status-service"
+    (register-status this status-core/status-service-name
                      status-core/status-service-version
                      1
                      (partial status-core/v1-status))
@@ -31,6 +31,10 @@
     (let [path (get-route this)
           handler (status-core/build-handler path (deref (:status-fns context)))]
       (add-ring-handler this handler))
+    context)
+
+  (stop [this context]
+    (status-core/reset-status-context! (:status-fns context))
     context)
 
   (register-status [this service-name service-version status-version status-fn]
--- a/test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
+++ b/test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
@@ -3,8 +3,8 @@
             [cheshire.core :as json]
             [schema.test :as schema-test]
             [puppetlabs.http.client.sync :as http-client]
-            [puppetlabs.trapperkeeper.services :refer [defservice service]]
-            [puppetlabs.trapperkeeper.app :refer [get-service]]
+            [puppetlabs.trapperkeeper.services :refer [defservice service service-context]]
+            [puppetlabs.trapperkeeper.app :refer [get-service] :as tka]
             [puppetlabs.trapperkeeper.testutils.bootstrap :refer :all]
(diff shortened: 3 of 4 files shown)


src/puppetlabs/trapperkeeper/services/status/status_service.clj
```diff
@@ -33,6 +33,11 @@
       (add-ring-handler this handler))
     context)
 
+  (stop [this context]
+    (status-core/dissoc-status-context! (:status-fns context)
+                                        status-core/status-service-name)
+    context)
```
cprice404, line 38: Couple of things here:
1. Is there any reason not to just clear the whole atom here?
2. The restart logic in TK involves throwing away the service's context entirely and replacing it with a clean one when it starts the service back up.  So I'm curious what kind of code you were running where that didn't sufficiently clear things out?

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -406,17 +406,28 @@
...
-       (is (= 200 (:status resp)))
-       (let [body (parse-response resp true)]
-         (is (= {:detail_level "debug"
-                 :service_name "status-service"
-                 :service_status_version 1
-                 :service_version status-core/status-service-version
-                 :state "running"}
-                (dissoc body :status)))
-         (is (map? (get-in body [:status :experimental :jvm-metrics]))))))))
+      app []
```
cprice404: I don't care about this too much but it seems like you are using 2-space indent, and that is probably going to cause whitespace flip-flopping when other people come in behind you?

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -406,17 +406,28 @@
...
+      (testing "and does so idempotently"
+        (let [get-status-status #(-> (get-service app :StatusService)
+                                     service-context
+                                     :status-fns
+                                     deref
+                                     (get status-core/status-service-name))]
+          (is (not (nil? (get-status-status))))
+          (tka/stop app)
+          (is (nil? (get-status-status)))
+          (tka/start app)
```
cprice404: TK would never call start here w/o calling init first; probably best for this test to use `restart` instead.  But... I also think that doing that would make this test pass even w/o the new logic you've added to the `stop` fn.

vilmibm: yeah, i'm specifically mirroring what pe-file-sync does.

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
vilmibm: oops, should've noticed that. i will go back to rome.

src/puppetlabs/trapperkeeper/services/status/status_service.clj
vilmibm, line 38: 1. i'm totally fine with that, was just trying to do the simplest change first
2. `(tka/stop app) (tka/start app)`

cprice404, line 38: re: 2., that kind of seems like a bug in the file sync test then (or wherever you got that from).  I suspect that if you change that to 'restart' your test will stop failing, and that would unblock you on that stream of work.

That's not to say that this PR isn't still a good idea, but it would no longer be a blocker, AFAIK.

vilmibm, line 38: the tests are structured that way for a reason; it's doing testing between the file sync client and storage services and needs to make the storage service unreachable momentarily. It's possible to structure that test in another way but it's more than just switching to `restart`.

haus:
this looks reasonable to me.

MSLilah:
This looks reasonable to me as well, :+1:

vilmibm:
that failure is the usual intermittent tk-status failure.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 0.3.3
+
+This is a maintenance / bugfix release.
+
+* Allow trapperkeepr's `stop` and `start` functions to be called on the status
```
haus: keepr => keeper

PR #48, opened 2016-04-21
(TK-214) Replace ringutils with common middleware helpers

justinstoller:
Previously we copied and pasted (and slightly tweaked) the contents of
p.t.s.s.ringutils across our projects.

Now `puppetlabs/ring-middleware` contains the canonical versions of
those utils.

This patch replaces usage of the vendored ringutils with usage of the
commensurate functions from ring-middleware.

This PR also updates the text expectations as the common utilities use
a slightly different error return format.


justinstoller:
NB: this shouldn't be merged in its current state. It is blocked on merging the ringutils code into ring-middleware. I'm opening this PRs so folks can see the effects of the consolidation. Once the updated ring-middleware is released this PR will need to be updated to pull in the new version of that library this PR will fail travis until then, it does allow those interested to test against a checkout of ring-middleware though.

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -263,24 +263,24 @@
     (testing "returns a 400 when an invalid level is queried for"
       (let [resp (http-client/get "http://localhost:8180/status/v1/services?level=bar")]
         (is (= 400 (:status resp)))
-        (is (= {"type" "request-data-invalid"
-                "message" "Invalid level: :bar"}
+        (is (= {"error" {"type" "request-data-invalid"
+                         "message" "Invalid level: :bar"}}
```
KevinCorcoran, line 267: This change might be problematic if downstream projects are relying on this specific existing map structure.

@nathanielksmith thoughts?

KevinCorcoran:
👍 except for my one question about the error format.

justinstoller:
I'd like to update this with the [bump to 0.3.1](https://github.com/puppetlabs/ring-middleware/pull/32) if it gets merged prior to this.

Regardless, @nathanielksmith do you have thoughts on the error format?

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
vilmibm, line 267: I feel like the next release of tk-status should indicate a breaking change and this should be fine?

vilmibm:
:+1:, with the understanding that we have confidence that downstream people will notice that the tk-status release containing this change is breaking (...probably in specific contrived testing scenarios, but still)

KevinCorcoran:
works for me, 🚢 🇮🇹

KevinCorcoran:
... but we should at least remember to note that in the release notes / CHANGELOG

PR #49, opened 2016-04-25
(TK-298) Bring tk-status docs up to date

vilmibm:
This PR:
- Updates the documentation index
- Adds information on the simple endpoints
- Links to the status proxy docs

### DIFF
--- a/documentation/developers.md
+++ b/documentation/developers.md
@@ -31,7 +31,6 @@ Code sample:
     context))
 ```
 
-
 ## Implementing Your Status Function
 
 The `puppetlabs.trapperkeeper.services.status.status-core` namespace contains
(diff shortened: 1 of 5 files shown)


documentation/developers.md
```diff
@@ -56,6 +55,31 @@ might be implemented to utilize the `compare-levels` function:
...
+
+web-router-service: {
+  "puppetlabs.trapperkeeper.services.status.status-service/status-service": {
+    route: /status
+    server: default
+  }
+}
+```
+
+For information on HTTP serving of the status endpoint, see the [Status Proxy documentation](./status-proxy-service.md).
```
KevinCorcoran: I think this could probably be reworded to be a little clearer.  As it stands, my first thought is "well, I could just change the jetty configuration above to accept plaintext connections".  So maybe instead of "HTTP serving of the status endpoint" something like "enabling proxying of plaintext HTTP `/status` requests to an otherwise HTTPS-protected server" (or something less awkward).

documentation/wire-formats.md
```diff
@@ -66,7 +68,37 @@ of error that occurred.  For example, "service-status-version-not-found".
...
+
+will be returned.
+
+It is important to note that when using the simple endpoints, a 503 response
+indicates a status other than _running_ and not a problem with the status
+service itself. However, a 5xx other than 503 would indicate such.
+
+#### Encoding
+
+The content type for these endpoints is `text/plain; charset=utf-8`.
```
KevinCorcoran, line 104: This is quite redundant with the file above.

documentation/query-api.md
```diff
@@ -212,3 +214,65 @@ version and a specific detail level:
...
         }
     }
+
+
+## Simple (plaintext) Endpoints
+
+These endpoints are designed for load balancers that don't support any kind of
+JSON parsing or query parameter use. They return simple string bodies (either
+the state of the service in question or a simple error message) and a status
+code relevant to the status result.
```
rlinehan, line 224: I think it might be useful/help in clarifying if you say something like if you need non-https for your load balancer, you should use the proxy service with this.

documentation/wire-formats.md
vilmibm, line 104: Agreed, but I was following the established form in both files. Doesn't seem too bad to have it in both places

documentation/query-api.md
vilmibm, line 224: Updated.

KevinCorcoran:
👍 

@rlinehan any more feedback on this?

rlinehan:
Nope, lgtm, :shipit: .

PR #50, opened 2016-05-24
(SERVER-1310) Use standard error format

justinstoller:
This moves us to the standard error format described in the [nogui](https://github.com/puppetlabs/puppet-nogui/blob/master/patterns/api_style_guide.md#errors) repo.

~~This is currently Do Not Merge [DNM] prep for after we have released ring-middleware 0.4.0.~~
We will also need to:
- [x] Update this to pull in the newest ring-middleware
- [x] Potentially update this for the reorg of helpers in ring-middleware into a utils namespace (not sure if that will land in 0.4.0).

### DIFF
--- a/test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
+++ b/test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
@@ -201,8 +201,8 @@
     (testing "returns a 404 for service not registered with the status service"
       (let [resp (http-client/get "http://localhost:8180/status/v1/services/notfound")]
         (is (= 404 (:status resp)))
-        (is (= {"type" "service-not-found"
-                "message" "No status information found for service notfound"}
+        (is (= {"kind" "service-not-found"
+                "msg" "No status information found for service notfound"}
               (parse-response resp)))))))
 
 (deftest status-code-test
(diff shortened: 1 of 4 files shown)


test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -260,28 +260,29 @@
...
-              (parse-response resp)))))
-    (testing "returns a 400 when a non-existent status-version is queried for"
-      (let [resp (http-client/get (str "http://localhost:8180/status/v1/"
-                                    "services/foo?service_status_version=3"))]
-        (is (= 400 (:status resp)))
-        (is (= {"error" {"type"    "service-status-version-not-found"
-                         "message" (str "No status function with version 3 "
-                                     "found for service foo")}}
-              (parse-response resp)))))))
+    (with-test-logging
```
KevinCorcoran, line 263: Why was it necessary to add this usage of `with-test-logging`?

justinstoller, line 263: These tests create log messages that clutter up the test output without this test.

That was missed when we moved to use the ring-middleware wrapper previously. While updating these tests I figured it would be good to clean up the log output. I can pull it out, or move it into it's own commit. I should have at least put it into its own commit.

PR #52, opened 2016-06-01
(maint) Resolve transitive dependency conflict with slf4j-api

justinstoller:
This is blocking the release of 0.3.4 (do not know if this means we need to burn that tag).

Our release job is failing with:

```
Possibly confusing dependencies found:
[puppetlabs/trapperkeeper "1.4.0" :exclusions [org.clojure/clojure]] -> [ch.qos.logback/logback-classic "1.1.3"] -> [org.slf4j/slf4j-api "1.7.7"]
 overrides
[puppetlabs/ring-middleware "1.0.0" :exclusions [org.clojure/clojure]] -> [puppetlabs/http-client "0.5.0"] -> [org.slf4j/slf4j-api "1.7.13"]

Consider using these exclusions:
[puppetlabs/ring-middleware "1.0.0" :exclusions [org.clojure/clojure org.slf4j/slf4j-api]]

Aborting due to :pedantic? :abort
Build step 'Execute shell' marked build as failure
```

IIUC, we only see this dependency conflict when we go to release because because one of our testing dependencies (puppetlabs/http-client) solves the conflict for us during normal development.

Tested with `lein with-profile -dev deps` (sorry, didn't realize I should be doing this before).

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -23,6 +23,7 @@
                  [org.clojure/tools.logging "0.3.1"]
                  [org.clojure/tools.reader "1.0.0-alpha1"]
                  [org.clojure/tools.trace "0.7.9"]
+                 [org.slf4j/slf4j-api "1.7.13"]
                  ;; end list of version conflict resolution dependencies
 
                  [cheshire "5.6.1"]


KevinCorcoran:
Wow, that's terrible.  This is the first time I've heard of a such a thing happening.  I'm sure nobody regularly runs your `lein with-profile -dev deps` so I guess we've just been getting lucky.  This seems like a fine fix for now but it'd sure be great if we had a better story about how to avoid hitting this in the future.  @cprice404 any thoughts?

pedantic?--

And yeah, I think it's probably best to just burn that tag, since presumably there is no 0.3.4 jar.

cprice404:
Agree re: burning the tag - just mention it in the CHANGELOG.  I don't think devs should need to run that `with-profile` command either, but we could/should add that command to the beginning of the release job, and then the release job would fail before it created the tag.

PR #54, opened 2016-06-28
(PE-15985) Add alerts to status response

lindboe:
This PR enables developers to add a new key to their status callback
function's response, :alerts, where they can provide alert objects with
a severity and prose message. They will be returned by the HTTP endpoint
as a list under the key "active_alerts".

Earlier discussions indicated that alerts should only be shown at info
level, but typically services have determined what information to show
at which level, so I wrote this code to enable that, along with
documentation specifying that typically alerts should be shown at info
or below.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/status/status_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_core.clj
@@ -23,9 +23,14 @@
 (def State
   (schema/enum :running :error :starting :stopping :unknown))
 
+(def Alert
+  {:severity (schema/enum :error :warning :info)
+   :message schema/Str})
+
 (def StatusCallbackResponse
   {:state State
-   :status schema/Any})
+   :status schema/Any
+   (schema/optional-key :alerts) [Alert]})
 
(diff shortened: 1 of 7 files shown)


src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -299,12 +303,16 @@
          data (:status callback-resp)
          state (if-not (schema/check State (:state callback-resp))
                  (:state callback-resp)
-                 :unknown)]
+                 :unknown)
+         alerts (if (:alerts callback-resp)
```
mullr: `if-let` or `or` can dry this up a bit

mullr: or just `get` with a default

mullr:
Minor style note, otherwise :+1: from me

mullr:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

mullr:
Ignore that comment I deleted, overzealous command line tooling gone awry

lindboe:
Updated

PR #56, opened 2016-07-15
(PE-16504) Timeout query parameter

aperiodic:
Change all of the status endpoints to accept a "timeout" query
parameter, which should be the timeout value in integral seconds.
If the timeout value is not provided, the default value will be whatever
`status-core/check-timeout` returns for the check's level. Any timeout
value that's not a positive integer will cause a 400 to be returned to
the client.

Add a new arity to the `get-status` method of the StatusService
protocol. This arity takes a timeout argument that overrides the default
timeout value for the level.

Increase the default :critical level timeout to 30 seconds.

Interspersed with this are some unrelated improvements in the 
`(maint)` commits; see the messages for more details.

### DIFF
--- a/documentation/query-api.md
+++ b/documentation/query-api.md
@@ -42,6 +42,14 @@ This will return status information for all registered services in an applicatio
 all registered services will be provided at the requested level of detail.  If
 not provided, the default level is `"info"`.
 
+* `timeout`: Optional. An integer specifying the timeout for the check in seconds.
+  If not provided, the default timeout will depend on the level.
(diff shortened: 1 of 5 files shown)


jonathannewman:
Changes look fine to me.   

Would it be possible to add an update to the docs explaining the new param?  
https://github.com/puppetlabs/trapperkeeper-status/blob/master/documentation/query-api.md#get-statusv1services

jonathannewman:
Great addition!

PR #57, opened 2016-07-26
(maint) fix timeout in docs

steveax:
Prior to this commit, the docs erroneously indicated that the timeout
for info level was 30s rather than the correct 60s value.


libbymo:
LGTM!

PR #58, opened 2016-09-13
(TK-400) Add setting to log status data periodically

jpinsonault:
This commit adds functionality to periodically log status endpoint data
to a file.

It logs directly through logback, so a corresponding logger must be created in
whatever project is consuming tk-status, otherwise it will clutter up the main
log files


jpinsonault:
ping @cprice404 
Does this seem reasonable?

documentation/status-logging.md
```diff
@@ -0,0 +1,54 @@
...
+
+To take advantage of this you must do two things:
+
+1. Enable logging though you TK app's configuration. For example, to log every 2 seconds, your configuration might look like this:
+
+    ```
+    status: {
+        logging-frequency: 2000,
+        status-logging-enabled: true
+    }
```
cprice404, line 14: This is net new configuration, right?  This is the first thing that is configurable about the status service?

Might want to have a quick brainstorm about the section name if that's the case.  I haven't thought of anything obviously better than `status`, but, just a good thing to spend a few minutes mulling over.

We could also consider getting rid of the 'enabled' setting and just say that if `logging-frequency` (or `status-log-interval` or something) was set to a non-zero value then that implicitly enables it?  Not 100% sure if that's better, just food for thought.

src/puppetlabs/trapperkeeper/services/status/status_logging.clj
```diff
@@ -0,0 +1,22 @@
...
+  (:import (clojure.lang IFn)
+           (java.util.concurrent Future)))
+
+(schema/defn start-background-task :- Future
+  "Creates a future which calls callback every frequency milliseconds"
+  [frequency :- schema/Int
+   callback :- IFn]
+  (future (while true
+            (callback)
+            (Thread/sleep frequency))))
```
cprice404: It would be better to use tk-scheduler for this, even if it's a net-new dependency.  I have another feature for tk-status that I did a POC on - which calculates CPU usage and GC CPU usage in the background, and will require a scheduled thread as well, so I think that's something that will come into this repo eventually one way or another, so might as well do it now.

src/puppetlabs/trapperkeeper/services/status/status_logging.clj
```diff
@@ -0,0 +1,22 @@
...
+
+(defn logging-fn
+  "Log status information at the debug level as json"
+  []
+  (let [status (status-core/v1-status :debug)]
```
cprice404: I notice you're logging at debug here.  That seems like the best level if it works... but... have you tested this with your `<root>` logger configured to log at DEBUG level?  And validated that these messages don't show up in the root logger?  Just wanting to make sure that they weren't being filtered by the level.

cprice404: Derp - I'm an idiot, that's not a logging level above.  So the actual logging level is `info`, below.  In that case it seems like you definitely would have seen these messages in the root logger if that config wasn't working the way we wanted?  So, that's good?  I do think you probably want to log these at `debug`, though?

cprice404:
looks good, just a few discussions to have around the config syntax, and the consideration about switching to tk-scheduler.

documentation/status-logging.md
jpinsonault: Yeah this is new. For sure, I just figured I'd use it to get started.

Interval, that's the word I was looking for.

I'd be fine either way with regards to the "enabled" setting

src/puppetlabs/trapperkeeper/services/status/status_logging.clj
jpinsonault: Cool, that's exactly what I wanted

src/puppetlabs/trapperkeeper/services/status/status_logging.clj
jpinsonault: I believe that's correct, that we wouldn't see them. I'll test it out though. Yeah it probably makes more sense to log them at `debug`

jpinsonault:
Pushed a commit that switches to using tk-scheduler. Still an open question about the `status-logging-enabled` setting being needed or not, and if the tk config section should be named `status` or something else. 

Also I'll fix the travis failure

documentation/status-logging.md
cprice404: I guess I have a slight preference for ditching the `enabled` setting, keeping the section called `status` (or, alternately, `status-service`, and I don't know how to decide between those :( ).  And maybe I'd call the interval setting `debug-logging-interval`.

Man I hate naming things.  @rlinehan do you have any opinions on this?  You're always good at catching the most idiotic things about my naming suggestions :)

documentation/status-logging.md
```diff
@@ -0,0 +1,54 @@
+# Status logging
+
+TK-status can periodically log status data to a file in JSON format.
+
+To take advantage of this you must do two things:
+
+1. Enable logging though you TK app's configuration. For example, to log every 2 seconds, your configuration might look like this:
```
rlinehan: s/Enable logging though/Enable logging th_r_ough/

documentation/status-logging.md
rlinehan: Most of our other services have config settings that are `metrics` or `webrouting` or whatever, without `-service`, so just `status` seems the most reasonable.

I'm more in favor of no `enabled` setting - it seems like that will just end up making configuring it more complicated. Are there ever any other settings that the logging stuff would ever have? Would we ever have something about what status endpoints are logged or something? If we do think that we would ever add any logging-related settings, then maybe have like:

```
status: {
    debug-logging: {
         interval-seconds: 60
    }
}
```

or something like that?

I think I like calling it `debug-logging` over `status-logging`. I think.

jpinsonault: Those names sound good to me

I kind like your layout @rlinehan regardless of if the settings might expand. Seems cleaner to separate it out like that

cprice404: @rlinehan 's suggestion does seem the most future-proof.  Maybe there's a small tradeoff in complexity but I kind of think it's worth it since it doesn't seem out of the realm of possibility that we might add another setting there in the future, e.g. a list of services to log the status from, rather than all services.

So I guess what I'm saying is I'm on board with @rlinehan 's suggestion if it seems good to you, @jpinsonault

jpinsonault:
Pushed some more commits addressing the config format/naming, and added fixed the failing tests since the scheduler service was introduced

documentation/status-logging.md
```diff
@@ -0,0 +1,54 @@
...
+
+To take advantage of this you must do two things:
+
+1. Enable logging through you TK app's configuration. For example, to log every 2 seconds, your configuration might look like this:
+
+    ```
+    status: {
+        status-logging-interval: 2000,
+        status-logging-enabled: true
+    }
```
cprice404, line 14: This is out of date now, correct?

src/puppetlabs/trapperkeeper/services/status/status_debug_logging.clj
```diff
@@ -0,0 +1,11 @@
+(ns puppetlabs.trapperkeeper.services.status.status-debug-logging
+  (:require [clojure.tools.logging :as log]
+            [schema.utils :refer [validation-error-explain]]
+            [cheshire.core :as json]
+            [puppetlabs.trapperkeeper.services.status.status-core :as status-core]))
+
+(defn log-status
+  "Log status information at the debug level as json"
+  []
+  (let [status (status-core/v1-status :debug)]
```
cprice404: I wonder if it would be worth adding a fn to core called, e.g., `status-latest-version` or `latest-version-status`, that would just call into `v1-status` for now.  Then, you could call `status-latest` from here, and if we ever introduce a `v2-status`, it would be fairly obvious that that new fn needed to be updated to call the newer version.  Otherwise a maintainer would need to know to come in here and change this line to the later version as well?

Or, maybe better, the core namespace could contain a fn like this:

``` clj
(defn status-map-for-version [version level]
   (cond version
      nil? (v1-status level)
      ;; 2 (v2-status level ;; future
      1 (v1-status level)))
```

wdyt?

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -174,6 +176,7 @@
       app
       [jetty9-service/jetty9-service
        webrouting-service/webrouting-service
+       scheduler-service/scheduler-service
        status-service]
       (merge status-service-config
         {:web-router-service {:puppetlabs.trapperkeeper.services.status.status-service/status-service "/alternate-status"}})
```
cprice404, line 191: Do you think it would be manageable to add an assertion into one or more of these tests to validate that the logging was actually happening if the config is set up properly?

I was going to suggest maybe also adding a test to validate that everything starts up properly if the new config is missing, but this test probably implicitly covers that for now?

cprice404 (commented):
LGTM; minor docs issue, minor suggested refactor for easier future maintenance, and curious whether it would be sane to add a test for the new functionality.  Otherwise, +1

src/puppetlabs/trapperkeeper/services/status/status_debug_logging.clj
jpinsonault: I like your `status-latest-version` idea because it's obvious that it should be updated. `status-map-for-version` might be a good idea regardless, but it seems like you'd still have the problem of forgetting to update this function to pass `2` to `status-map-for-version`

I'll go with `status-latest-version` for now I think?

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
jpinsonault: Maybe, you'd have to set the interval low and then wait a small amount of time before checking I think, might be a little racy

Yeah it doesn't seem worth testing that it works without the config since that's covered implicitly and it seems pretty low risk

jpinsonault: Thinking about it more, I'm not sure there is a sane way to test it, since the log message will just be a json string, it'll be hard to call `logged?` with some regex. We could redef the logging function with a promise or something to make it more synchronous, but then we're just testing tk-scheduler, not that the service logs correctly ¯\_(ツ)_/¯

jpinsonault: Well that'd at least be testing that the config gets plumbed through...

cprice404: @jpinsonault you could use this:

https://github.com/puppetlabs/trapperkeeper/blob/master/test/puppetlabs/trapperkeeper/testutils/logging.clj#L210-L220

Then you'd actually be able to de-json the string from the log message.

And I don't think the test would necessarily be racey as long as you slept for some interval longer than the logging interval before looking for the message, and tolerated more than one message in the log output in case the scheduler had fired twice.

That said, I agree it's kind of gross, and could live with the argument that it's not worth testing.  Maybe @rlinehan or @camlow325 has an opinion?

jpinsonault:
Pushed a few more commits addressing the latest comments. Went ahead and added a test to make sure logging happens. Not too weird looking

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -80,6 +80,9 @@
    :up-time-ms WholeMilliseconds
    :start-time-ms WholeMilliseconds})
 
+(def DebugLoggingConfig
+  (schema/maybe {:interval-milliseconds schema/Int}))
```
cprice404: Question - did you change this to millis mostly for the sake of the test?  I don't think a user would probably ever want to configure this for less than 1 second (it seems way more likely that it's going to be something like every 30 mins), and we should probably optimize the configuration for the user use case.

Do you know whether a float would make its way through the system?  If so, I might suggest that we change this to `interval-minutes`, and then for the test use a float < 1?

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -469,3 +478,35 @@
                  :state "running"}
                 (dissoc body :status)))
          (is (map? (get-in body [:status :experimental :jvm-metrics]))))))))
+
+(deftest status-debug-logging-test
+  (testing "status service logs debug data when setting is enabled"
+    (with-test-logging
+     (with-logger-event-maps
```
cprice404, line 485: Do you need `with-test-logging` if you are using `with-logger-event-maps`?  I thought maybe the `with-test-logging` would be unnecessary in that case, but not sure.

cprice404:
+1 other than the question about the time unit we use in the config, and the question about whether we still need `with-test-logging` for the new tests.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
jpinsonault: I changed the name because the value was in milliseconds before, it was just mislabeled. 

Yeah, agreed that the user wouldn't want milliseconds. We could change it to minutes, and just convert it to milliseconds for the scheduler service. Tried putting a float in, it works fine

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
jpinsonault: Yeah without `with-test-logging`, I wouldn't see any log messages at all

src/puppetlabs/trapperkeeper/services/status/status_core.clj
cprice404: Cool, let's change it to minutes then.

cprice404: And yeah it's fine if it's represented as millis internally.

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
cprice404: Gotcha.  This is fine, then.

jpinsonault:
Pushed a commit changing it to minutes

src/puppetlabs/trapperkeeper/services/status/status_service.clj
```diff
@@ -40,10 +40,11 @@
       (add-ring-handler this handler))
     (let [debug-logging-config (schema/validate status-core/DebugLoggingConfig
                                                 (get-in-config [:status :debug-logging]))
-          debug-logging-interval (:interval-milliseconds debug-logging-config)]
-      (when debug-logging-interval
-        (log/info "Starting background logging of status data")
-        (interspaced debug-logging-interval status-logging/log-status)))
+          interval-minutes (:interval-minutes debug-logging-config)]
+      (when interval-minutes
+        (let [interval-milliseconds (* 60000 interval-minutes)]
```
cprice404, line 45: This is totally fine like this, but just FYI, Java's built-in `TimeUnit` is pretty nifty for stuff like this:

``` clj
(.toMillis TimeUnit/MINUTES 30)
=> 1800000
```

Just something good to know about.

cprice404 (approved):
I'm +1 now.  @rlinehan @camlow325 if either of you get a minute, I think this is ready for a second review.

cprice404:
Also - I'd mentioned the possibility of holding off on merging this until after we roll out the logback changes, but I think I've drifted over to Jeremy's side of the fence on that one.  I think it's probably fine to merge these now.  We should be rolling out those logback changes soon-ish, and this feature is disabled by default, and even if the PR is merged this isn't in a release of tk-status yet... so I think I'm good with just getting it merged now.

src/puppetlabs/trapperkeeper/services/status/status_service.clj
jpinsonault: Oh nice, yeah good to know

documentation/status-logging.md
```diff
@@ -0,0 +1,55 @@
+# Status logging
+
+TK-status can periodically log status data to a file in JSON format.
+
+To take advantage of this you must do two things:
+
+1. Enable logging through you TK app's configuration. For example, to log every 2 seconds, your configuration might look like this:
```
camlow325: s/you/your

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
camlow325: @cprice404 The extra testing that you were suggesting is covered by the new `status-debug-logging-test` now, right?

src/puppetlabs/trapperkeeper/services/status/status_debug_logging.clj
```diff
@@ -0,0 +1,14 @@
...
+            [cheshire.core :as json]
+            [puppetlabs.trapperkeeper.services.status.status-core :as status-core]))
+(defn log-status
+  "Log status information at the debug level as json
+
+  Note: This function is in its own namespace so that logback can use the namespace
+  as a way to route these log messages separately from other logging the application
+  might be doing, and so it shouldn't be moved from this namespace"
+  []
+  (let [status (status-core/status-latest-version :debug)]
```
camlow325, line 13: @cprice404, @jpinsonault - might be nice to have the ability to use TK-level parameters when doing the status logging, e.g., to be able to select a higher or lower level of granularity when logging rather than just hard-coding it to "debug".  Suppose we could just add other namespaces like this one later if we find the need, e.g., a "status-trace-logging" (assuming the 'trace' level is a thing, can't remember).  Figured I'd just throw that out for thoughts but I'm happy going with what we have already here if ye are.

documentation/status-logging.md
camlow325: The units in the example here seem inconsistent.  The description says "to log every 2 seconds" but the HOCON snippet has `interval-minutes: 5`.  Should we change the text to say "to log every 5 minutes"?

documentation/status-logging.md
```diff
@@ -0,0 +1,55 @@
...
+            <totalSizeCap>1GB</totalSizeCap>
+        </rollingPolicy>
+        <encoder>
+            <!-- note that this will only log the JSON message (%m) and a newline (%n)-->
+            <pattern>%m%n</pattern>
+        </encoder>
+    </appender>
+
+    <!-- without additivity="false", the status log messages will be sent to every other appender as well-->
+    <logger name="puppetlabs.trapperkeeper.services.status.status-logging" additivity="false">
```
camlow325: These examples all have `status-logging` but the only namespace in the Clojure code that does this is `status-debug-logging`, right?  I can only get the logging to go to a file and not other loggers if I use:

``` xml
<logger name="puppetlabs.trapperkeeper.services.status.status-debug-logging" additivity="false">
...
```

documentation/status-logging.md
camlow325: Although I suppose it might be good to point out that you can use fractions of a minute if you really want more granularity, like .5 to achieve 30 seconds.

camlow325 (changes requested):
Overall, 👍 Left a few nits about the documentation.  The only more involved question I left was about whether we just wanted to have a "debug-logging" section or something a little more extensible like "logging" with a subkey of "level" where you could tune the level to "debug", "trace", "info", or whatever.  Happy to defer to ye all on that question though.

documentation/status-logging.md
cprice404: I don't think I'd document the float.  And I'd set the examples up to have a realistic value that a real-world user might actually use, like 30 mins or an hour or 2 hours.

camlow325: k, sounds fine to me.

documentation/status-logging.md
jpinsonault: oh good catch

src/puppetlabs/trapperkeeper/services/status/status_debug_logging.clj
jpinsonault: I'm personally fine with leaving it as only debug data for now. It seems like most users will want the debug level since that's where jvm metrics show up in the status service

jpinsonault:
Pushed a commit updating the docs

src/puppetlabs/trapperkeeper/services/status/status_debug_logging.clj
cprice404: Yeah, I don't have a super strong opinion one way or the other but this seems like it will be the common use case, so I'm good with it as-is unless someone feels strongly the other way.

camlow325: I don't have that strong of opinion either so I'm good just sticking with it as is.

camlow325 (approved):

camlow325:
:shipit:

PR #59, opened 2016-09-14
Add i18n library and mark strings up for string externalization

lutter:


lutter:
I just realized that I forgot to wrap service handlers in `i18n/locale-negotiator`. I _believe_ that the right place to do that would be `status-core/build-handler` but would prefer if somebody who understands the code in more detail confirmed that that's true.

cprice404:
@lutter I think you are correct, `build-handler` is the right place for that.

jpinsonault:
I've rebased and added the locale-negotiator middlware. Hope that works ¯\_(ツ)_/¯

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -460,15 +460,16 @@
...
-  (comidi/routes->handler
-   (comidi/wrap-routes
-    (comidi/context path
-      (comidi/context "/v1"
-        (-> (build-json-routes "/services" status-fns)
-            (comidi/wrap-routes (errors-by-type-middleware :json)))
-        (-> (build-plaintext-routes "/simple" status-fns)
-            (comidi/wrap-routes (errors-by-type-middleware :plain)))))
-    #(ring-defaults/wrap-defaults % ring-defaults/api-defaults))))
+  (i18n/locale-negotiator
```
cprice404: This _might_ be correct, but it's not what I was expecting it to look like.  I was expecting some i18n middleware to be added to the stuff in the `comidi/wrap-routes` call below.  e.g. composed with that anonomous fn to `wrap-defaults`.

If you haven't already, it's probably worth throwing together a little test setup where you print out some debugging info from inside the handler function... to print out the extra data that should be injected by that middleware and convince yourself that it's actually being set?  Would be awesome to write a test to cover it but it kind of feels like we can't do that without having some translation files.

jpinsonault: Yeah I'm not sure how to test this locally, but I'm attempting to setup a test environment and make sure it gets the right locale in requests

jpinsonault:
Ok, I managed to get a test setup where I logged the user locale, and confirmed that the middleware is setting the locale correctly. The trick was mostly that I needed to create a translation file for a language before it would acknowledge that I'd requested that locale in the middleware

lutter:
LGTM .. though I don't feel qualified to comment on whether this is the right place to put in `locale-negotiator` or not. If @cprice404  is happy with it, so am I.

cprice404:
@jpinsonault That looks more like what I was expecting.  If you've tested it, I'm good.

PR #62, opened 2016-10-02
(TK-409) Add gc counts and file descriptor usage to jvm-metrics

Sharpie:
This PR adds two new entries to the `jvm-metrics` section of status output:
- `file-descriptors`: Which includes the maximum number of file descriptors allowed and the count of currently used descriptors.
- `gc-stats`: An array containing the count of GC events and time spent on garbage collection for each collector.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/status/status_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_core.clj
@@ -1,5 +1,6 @@
 (ns puppetlabs.trapperkeeper.services.status.status-core
   (:require [clojure.tools.logging :as log]
+            [clojure.set :as setutils]
             [schema.core :as schema]
             [schema.utils :refer [validation-error-explain]]
             [ring.middleware.defaults :as ring-defaults]
@@ -74,9 +75,14 @@
    :max schema/Int
    :used schema/Int})
 
+(def FileDescriptorUsageV1
+  {:max schema/Int
+   :used schema/Int})
+
 (def JvmMetricsV1
   {:heap-memory MemoryUsageV1
    :non-heap-memory MemoryUsageV1
+   :file-descriptors FileDescriptorUsageV1
    :up-time-ms WholeMilliseconds
    :start-time-ms WholeMilliseconds})
 
@@ -152,6 +158,9 @@
   (let [runtime-bean (ManagementFactory/getRuntimeMXBean)]
     {:heap-memory (jmx/read "java.lang:type=Memory" :HeapMemoryUsage)
      :non-heap-memory (jmx/read "java.lang:type=Memory" :NonHeapMemoryUsage)
+     :file-descriptors (setutils/rename-keys
+                        (jmx/read "java.lang:type=OperatingSystem" [:OpenFileDescriptorCount :MaxFileDescriptorCount])
+                        {:OpenFileDescriptorCount :used :MaxFileDescriptorCount :max})
      :up-time-ms (.getUptime runtime-bean)
      :start-time-ms (.getStartTime runtime-bean)}))
 
--- a/src/puppetlabs/trapperkeeper/services/status/status_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_core.clj
@@ -79,10 +79,15 @@
   {:max schema/Int
    :used schema/Int})
 
+(def GcStatsV1
+  {schema/Str {:count schema/Int
+               :total-time-ms schema/Int}})
+
 (def JvmMetricsV1
   {:heap-memory MemoryUsageV1
    :non-heap-memory MemoryUsageV1
    :file-descriptors FileDescriptorUsageV1
+   :gc-stats GcStatsV1
    :up-time-ms WholeMilliseconds
    :start-time-ms WholeMilliseconds})
 
@@ -155,12 +160,19 @@
 
 (schema/defn ^:always-validate get-jvm-metrics :- JvmMetricsV1
   []
-  (let [runtime-bean (ManagementFactory/getRuntimeMXBean)]
+  (let [runtime-bean (ManagementFactory/getRuntimeMXBean)
+        gc-beans (jmx/mbean-names "java.lang:name=*,type=GarbageCollector")]
     {:heap-memory (jmx/read "java.lang:type=Memory" :HeapMemoryUsage)
      :non-heap-memory (jmx/read "java.lang:type=Memory" :NonHeapMemoryUsage)
      :file-descriptors (setutils/rename-keys
                         (jmx/read "java.lang:type=OperatingSystem" [:OpenFileDescriptorCount :MaxFileDescriptorCount])
                         {:OpenFileDescriptorCount :used :MaxFileDescriptorCount :max})
+     :gc-stats (into {} (for [gc gc-beans]
+                          (let [gc-name (.getKeyProperty gc "name")
+                                gc-info (setutils/rename-keys
+                                         (jmx/read gc [:CollectionCount :CollectionTime])
+                                         {:CollectionCount :count :CollectionTime :total-time-ms})]
+                            {gc-name gc-info})))
      :up-time-ms (.getUptime runtime-bean)
      :start-time-ms (.getStartTime runtime-bean)}))
 
--- a/test/puppetlabs/trapperkeeper/services/status/status_core_test.clj
+++ b/test/puppetlabs/trapperkeeper/services/status/status_core_test.clj
@@ -110,10 +110,13 @@
       (is (= #{:jvm-metrics} (ks/keyset (get-in status [:status :experimental]))))
       (let [jvm-metrics (get-in status [:status :experimental :jvm-metrics])]
         (is (= #{:heap-memory :non-heap-memory
+                 :file-descriptors
                  :up-time-ms :start-time-ms} (ks/keyset jvm-metrics)))
         (is (= #{:committed :init :max :used} (ks/keyset (:heap-memory jvm-metrics))))
         (is (= #{:committed :init :max :used} (ks/keyset (:non-heap-memory jvm-metrics))))
         (is (every? #(< 0 %) (vals (:heap-memory jvm-metrics))))
         (is (every? #(or (< 0 %) (= -1 %)) (vals (:non-heap-memory jvm-metrics))))
+        (is (= #{:max :used} (ks/keyset (:file-descriptors jvm-metrics))))
+        (is (every? #(< 0 %) (vals (:file-descriptors jvm-metrics))))
         (is (< 0 (:up-time-ms jvm-metrics)))
         (is (< 0 (:start-time-ms jvm-metrics)))))))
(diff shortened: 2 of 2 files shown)


src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -149,9 +161,17 @@
...
+        gc-beans (jmx/mbean-names "java.lang:name=*,type=GarbageCollector")]
     {:heap-memory (jmx/read "java.lang:type=Memory" :HeapMemoryUsage)
      :non-heap-memory (jmx/read "java.lang:type=Memory" :NonHeapMemoryUsage)
+     :file-descriptors (clojure.set/rename-keys
+       (jmx/read "java.lang:type=OperatingSystem" [:OpenFileDescriptorCount :MaxFileDescriptorCount])
+       {:OpenFileDescriptorCount :used :MaxFileDescriptorCount :max})
+     :gc-stats (map #(conj {:name (.getCanonicalName %)}
+       (clojure.set/rename-keys (jmx/read % [:CollectionCount :CollectionTime])
+         {:CollectionCount :count :CollectionTime :total-time-ms}))
+       gc-beans)
```
jpinsonault: This might be easier to parse if the arguments to map and the lambda in the map had their contents aligned. Hard to see easily what is an argument to what

jpinsonault:
I'm 👍 on this, just had a small style suggestion

src/puppetlabs/trapperkeeper/services/status/status_core.clj
Sharpie: Pushed an updated patch that is formatted according to the suggestions of `lein cljfmt`. Let me know if it still needs tweaking!

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -149,9 +161,17 @@
 
 (schema/defn ^:always-validate get-jvm-metrics :- JvmMetricsV1
   []
-  (let [runtime-bean (ManagementFactory/getRuntimeMXBean)]
+  (let [runtime-bean (ManagementFactory/getRuntimeMXBean)
+        gc-beans (jmx/mbean-names "java.lang:name=*,type=GarbageCollector")]
     {:heap-memory (jmx/read "java.lang:type=Memory" :HeapMemoryUsage)
      :non-heap-memory (jmx/read "java.lang:type=Memory" :NonHeapMemoryUsage)
+     :file-descriptors (clojure.set/rename-keys
```
cprice404: Nit: we try to avoid using fully qualified namespaces in the code because they can allow situations where tests pass even though you haven't actually done the necessary `require` in the namespace, if it happened to get `required` in some other namespace.  This can lead to ordering issues where a namespace works fine if loaded _after_ some other namespace, but doesn't work on its own.  The preferred approach is to add an `:as` clause to your `require` (e.g., `:as set`), and then use that alias in the code.

Sharpie: I thought about that, but `:as set` seems like it might shadow `clojure.core/set`. Couldn't think of a better alias, so I just went with the namespace since it's pretty short to begin with.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -149,9 +161,17 @@
...
+                        (jmx/read "java.lang:type=OperatingSystem" [:OpenFileDescriptorCount :MaxFileDescriptorCount])
+                        {:OpenFileDescriptorCount :used :MaxFileDescriptorCount :max})
+     :gc-stats (map #(conj {:name (.getCanonicalName %)}
```
cprice404: Do you expect the resulting list of metrics here to be deterministic?  Or is it possible that it will change depending on which GC implementation is in use?  I'm just thinking about trying to document this as an API, and whether or not we can make guarantees about the keys that will/won't exist.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -149,9 +161,17 @@
...
+                           (clojure.set/rename-keys (jmx/read % [:CollectionCount :CollectionTime])
+                                                    {:CollectionCount :count :CollectionTime :total-time-ms}))
+                    gc-beans)
```
cprice404: I need to pull this down and run it to confirm, but if I'm reading this correctly, the value for the `:gc-stats` entry in the status results will be a sequence/list, as opposed to a map/object.  That means that people would need to write code to iterate over that list (or assume they can call `first` on it) in order to consume this data.  Would prefer to see if we can figure out how to get this data converted into a more predictable, deterministic map structure.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
Sharpie: Looks like CollectionCount and CollectionTime are two attributes that _must_ be present if a MBean implements `java.lang:type=GarbageCollector`:

https://docs.oracle.com/javase/8/docs/api/java/lang/management/GarbageCollectorMXBean.html

src/puppetlabs/trapperkeeper/services/status/status_core.clj
cprice404: Yeah, the length isn't the issue - it's just about making sure you refer to an alias rather than the fully qualified namespace.  You could use `:as clojure-set` if you prefer.

It won't shadow `clojure.core/set`, though, because that's a function/var, and this is a namespace alias, and clojure treats those differently.  But I've no problem with using a less ambiguous alias if you prefer.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
cprice404: Is it possible for there to be more than one such MBean in a process?

src/puppetlabs/trapperkeeper/services/status/status_core.clj
Sharpie: Yup. It currently returns a list of `[{:name, :count, :total-time-ms}]`. Would be great to consolidate that into a map like `{:young-gc {...}, :tenured-gc {...}}`. But:
- We'd need to find or maintain a mapping of GC algorithm names -> tenured/young.
- This assumes there's always two GC algorithms. This appears to be the case for the default Scavange/MarkSweep and G1GC, but I'm not sure if it holds for all GC options.

Looks like Dropwizard Metrics returns a map of GC algorithm name -> metric set:

https://github.com/dropwizard/metrics/blob/v3.1.2/metrics-jvm/src/main/java/com/codahale/metrics/jvm/GarbageCollectorMetricSet.java

Could look into doing something similar.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
Sharpie: There will be one per collection algorithm. Not sure if it is possible to create two instances of the same algorithm.

cprice404: I kind of like the approach you linked from Dropwizard.  Or we could do something somewhat similar where we use the algorithhm name as a key in a map, and then put the values into a nested map.  Then the json traversal becomes deterministic;  you'd need to know the names of the GC algos that you were running in your environment, but you wouldn't have to iterate over an array.

cprice404:
I think I'm basically +1 on this, pending outcome of discussion on how to organize the GC metrics, and assuming we don't believe there would be any downside to having this available alongside the data described in [PE-13346](https://tickets.puppetlabs.com/browse/PE-13346) (which I would really like to include in a graph in the clojurescript developer dashboard).

@rlinehan you might be interested in checking this PR out if you get time.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
Sharpie: Alright, I'll take a shot at `{"<collector name>" {:count :total-time-ms} ...}`.

Right now, the schema for the list is:

```
(def GcStatsV1
  [{:name schema/Str
    :count schema/Int
    :total-time-ms schema/Int}])
```

That would need to change to something like:

```
(def GcStatsV1
  {(schema/OneOrMore schema/Str) {:count schema/Int
                                  :total-time-ms schema/Int}})
```

The `schma/OneOrMore` above is just a placeholder I made up --- what would be the proper Schema definition for repeated map entries which follow a particular pattern?

jpinsonault: I think that'd just be 

```
(def GcStatsV1
  {schema/Str {:count schema/Int
               :total-time-ms schema/Int}})
```

Sharpie: That works!

Sharpie:
PR updated to address:
- Aliasing `clojure.set`
- `gc-stats` is now a map instead of a list. The GC algorithm names are used as the keys of this map and the values are a sub-map containing the collection counts and total time spent on collections.

cprice404:
I've pulled this down and run it locally with Puppet Server.  LGTM.  We need to add some docs about the metrics, but it appears that I did not do that when I added the original metrics, so there's no place to add these fields to at the moment.  I'll throw up a separate PR for docs.

+1 from me.  @rlinehan after I get that docs PR up you might want to look at this and see if you have any comments on the data structure.

cprice404:
Docs PR: #63

test/puppetlabs/trapperkeeper/services/status/status_core_test.clj
```diff
@@ -110,10 +110,18 @@
...
         (is (= #{:committed :init :max :used} (ks/keyset (:heap-memory jvm-metrics))))
         (is (= #{:committed :init :max :used} (ks/keyset (:non-heap-memory jvm-metrics))))
         (is (every? #(< 0 %) (vals (:heap-memory jvm-metrics))))
         (is (every? #(or (< 0 %) (= -1 %)) (vals (:non-heap-memory jvm-metrics))))
+        (is (= #{:max :used} (ks/keyset (:file-descriptors jvm-metrics))))
+        (is (every? #(< 0 %) (vals (:file-descriptors jvm-metrics))))
+        (is (every? #(= #{:count :total-time-ms} (ks/keyset %))
+                    (vals (:gc-stats jvm-metrics))))
+        (is (every? #(<= 0 %) ;; Possible that no major collections occurred.
+                    (mapcat #(vals %)
```
rlinehan, line 124: A small clojure thing - this doesn't need to be `(mapcat #(vals %)...` - it can just be `(mapcat vals...` functions that take a single arg can be passed to mapping functions without parens. I think basically `map` and friends take as their first argument a function that takes a single argument. Wrapping things in `#()` or `(fn [x])` creates a function that takes one arg, but isn't necessary if you are using an already existing function that takes a single arg (this also includes keywords, since keywords are functions - e.g. `(map :a [{:a 1} {:a 2} {:a 3}]) => (1 2 3)`)

(Not something that needs to be fixed in this PR - just an FYI).

rlinehan:
@cprice Just so I understand, this commit adds counts and time for GC usage and counts of file descriptors, and PE-13346 would add CPU usage and CPU usage for GC? If that's the case, then yeah, I think both seem like good data to have, and I see no downside.

I'm 👍 on this. I like having the algorithm names as keys in a map, rather than a sequence of maps. It might be good to document what some of the options are for GC names somewhere, although maybe they're just readily known/google-able? But that's more a comment on the docs PR.

rlinehan:
Talked to @cprice404 in person. We decided that since this is already marked as `experimental` and since changes for PE-13346 to add CPU usage would be additive, we're good with merging this (and releasing it) now and can make those changes later.

tl, dr: :shipit:

PR #63, opened 2016-10-05
(MAINT) Add docs about metrics, JVM metrics

cprice404:


documentation/metrics.md
```diff
@@ -0,0 +1,78 @@
+## Metrics
+
+It is common for services registering callbacks with the status service to include some basic metrics about the health
+of the service.  These metrics are typically only available at the `:debug` status level, and should be limited to a
+fairly small set of data that can be useful for debugging the service's performance or behavior.
```
rlinehan: Might be good to give an example here?

rlinehan: Would it be good to mention that these have typically been labeled as "experimental"/how we want to version them?

documentation/metrics.md
```diff
@@ -0,0 +1,78 @@
...
+* `heap-memory`: information about the JVM's heap memory usage; this mostly accounts for memory consumed by application code
+** `committed`: the amount of memory that the operating system has allocated to the process
+** `init`: the initial amount of memory that was allocated to the process at startup
+** `max`: the maximum amount of memory that the process will request from the operating system before throwing an OOM error
+** `used`: the amount of memory that is currently being used by the application
+* `non-heap-memory`: same fields as for `heap-memory`, but this refers to native memory used by the JVM itself, along with any memory allocated by native libraries
+* `file-descriptors`:
+** `max`: the maximum number of file descriptors that this process is allowed to open
+** `used`: the current number of file descriptors the process has open
+* `gc-stats`: a map containing key garbage collection statistics for each of the different GC algorithms that are in use by this JVM.  The keys in this map are the GC algorithm names
```
rlinehan, line 77: Is there any documentation or anything that would be useful to link to as far as what these GC algorithm names are, or is that common enough knowledge?

cprice404: I've never found any really great links about this.  Here are the best couple that I've found, if anyone thinks they are relevant enough (and unlikely to be found by a user on their own), I can certainly add one or both of them:

http://docs.oracle.com/javase/8/docs/technotes/guides/vm/gctuning/collectors.html#sthref27
http://www.oracle.com/technetwork/java/javase/tech/gc-142724.html

rlinehan: Hmm, neither of those seem great. I guess probably people could google the names they see in the output and get better results. So I'm good with not including this.

rlinehan:
👍

documentation/metrics.md
cprice404: Added a commit that adds a bit more treatment here.  It links to the pe-pse repo as an example, which is unfortunately closed source, but seemed like the best example we have at the moment.  We could update it to point to the OSS repo when we move some of the metrics to OSS.

rlinehan:
💯 🚢 🍰

rlinehan (approved):

PR #64, opened 2016-10-05
(MAINT) Bump tk and http-client deps

camlow325:
This commit bumps the following deps:
- trapperkeeper - 1.4.0 -> 1.5.1
- clj-http-client - 0.5.0 -> 0.6.0

This commit also removes the explicit slf4japi dep since it would no
longer conflict between the trapperkeeper and clj-http-client deps.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(def tk-version "1.4.0")
+(def tk-version "1.5.1")
 (def ks-version "1.3.0")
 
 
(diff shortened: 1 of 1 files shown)


cprice404:
Would it be worth just going ahead and porting this over to `lein-parent` instead of doing this?

camlow325:
@cprice404 Yeah, I'm okay with that.  You want me to do that or was this one already on your list?

cprice404:
I  haven't done this one yet.  So if you need these changes for something else then you might want to go ahead and do this repo.  I just have tk, tk-j9, and puppetserver pending.

camlow325:
As it turns out, I don't need to do these version bumps yet, so I'll just close this one for now.  If I do come back around to this later, I'll do the dependency bumps via `lein-parent` / `clj-parent`.

PR #65, opened 2016-10-26
[DON'T MERGE YET] (maint) Update ring-middleware and http-client dep

briancain:
**This can't be merged until `ring-middleware 1.0.1` is released.**

Related PR: https://github.com/puppetlabs/ring-middleware/pull/40

If a trapperkeeper service brings in a previous version of trapperkeeper
status but is not using the clj-http-client project, starting up
tk-status results in a java.lang.ClassNotFound exception inside the
ring-middleware library. This commit updates the ring-middleware dep to
fix that problem so the class exists.

Stacktrace: https://gist.github.com/briancain/034f7ea9d2954f6d6f6f796095ec0016

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -43,7 +43,7 @@
                  [puppetlabs/kitchensink ~ks-version]
                  [puppetlabs/trapperkeeper ~tk-version]
                  [puppetlabs/trapperkeeper-scheduler "0.0.1"]
-                 [puppetlabs/ring-middleware "1.0.0"]
+                 [puppetlabs/ring-middleware "1.0.1" :exclusions  [org.clojure/clojure org.slf4j/slf4j-api]]
                  [puppetlabs/comidi "0.3.1"]
                  [puppetlabs/i18n "0.4.3"]]
 
@@ -61,7 +61,7 @@
                                   [commons-io "2.5"]
                                   ;; End transitive dependency resolution
 
-                                  [puppetlabs/http-client "0.5.0"]
+                                  [puppetlabs/http-client "0.6.0" :exclusions  [org.clojure/clojure org.slf4j/slf4j-api]]
                                   [puppetlabs/trapperkeeper ~tk-version :classifier "test"]
                                   [puppetlabs/trapperkeeper-webserver-jetty9 "1.5.6"]
                                   [puppetlabs/kitchensink ~ks-version :classifier "test"]]}}


project.clj
```diff
@@ -43,7 +43,7 @@
                  [puppetlabs/kitchensink ~ks-version]
                  [puppetlabs/trapperkeeper ~tk-version]
                  [puppetlabs/trapperkeeper-scheduler "0.0.1"]
-                 [puppetlabs/ring-middleware "1.0.0"]
+                 [puppetlabs/ring-middleware "1.0.1" :exclusions  [org.clojure/clojure org.slf4j/slf4j-api]]
```
rlinehan, line 46: Rather than doing `exclusions` like this we've been pinning the version in a "dependency conflict resolution" list - e.g. https://github.com/puppetlabs/trapperkeeper-status/pull/65/files#diff-0fff143854a4f5c0469a3819b978a483R27.

Probably this just means bumping the `slf4j-api` version in the conflict list?

rlinehan:
If I add tk-status on top of the [PR enabling pedantic on dujour](https://github.com/puppetlabs/dujour/pull/54/files), then I get a whole bunch of version conflicts:

```
$ lein deps
Possibly confusing dependencies found:

... (a whole bunch of other stuff)

[grimradical/ring-geoipviz "0.1.0" :exclusions [org.clojure/clojure ring/ring-core commons-fileupload commons-io]] -> [geocoder-clj "0.0.6"] -> [clj-http "0.5.5"] -> [org.apache.httpcomponents/httpclient "4.2.1"]
 overrides
[puppetlabs/trapperkeeper-status "0.5.0"] -> [puppetlabs/ring-middleware "1.0.0" :exclusions [org.clojure/clojure]] -> [puppetlabs/http-client "0.5.0"] -> [org.apache.httpcomponents/httpasyncclient "4.1.1"] -> [org.apache.httpcomponents/httpclient "4.5.1"]

Consider using these exclusions:
[puppetlabs/trapperkeeper-status "0.5.0" :exclusions [org.apache.httpcomponents/httpclient]]

Aborting due to :pedantic? :abort
```

I think that what's actually happening is that lein is choosing `org.apache.httpcomponents/httpclient "4.2.1"` from the `geocoder-clj` dependency, and 4.2.1 doesn't have the `RequestConfig` class. In http-client `0.6.0` this isn't a problem because we've gotten rid of the import that pulls in `RequestConfig`.

However, I think a better solution is to fix all the dependency conflicts in dujour when tk-status is introduced and you're using pedantic, preferably by the method described in https://confluence.puppetlabs.com/pages/viewpage.action?pageId=48956980. If I just add `[org.apache.httpcomponents/httpclient "4.5.2"]` to the dependency list in dujour in addition to adding `[puppetlabs/trapperkeeper-status "0.5.0"]` then `lein tk` works fine.

briancain:
thanks @rlinehan ! I'll go ahead and close the PRs.

PR #67, opened 2016-10-27
(TK-416) Add type and details to alerts.

adreyer:


documentation/developers.md
```diff
@@ -65,6 +65,42 @@ might be implemented to utilize the `compare-levels` function:
...
+readable information about the state of your application. This makes it easy
+for status to be displayed and checked by unopinionated consumers. This works
+best if all conditions which may effect the availability of your service
+result in an alert.
+
+Examples:
+```clj
+{:severity :error
+ :message "Cannot connect to database"
+ :type :my-service/databas-unavailable
```
rlinehan: s/databas/database/

documentation/developers.md
```diff
@@ -65,6 +65,42 @@ might be implemented to utilize the `compare-levels` function:
...
+ :details {}}
+
+{:severity :warning
+ :message "Repo may not be initialized"
+ :type :my-service/uninitilized-repo
```
rlinehan: s/uninitilized/uninitialized/

documentation/developers.md
```diff
@@ -65,6 +65,42 @@ might be implemented to utilize the `compare-levels` function:
...
+ :details { :repo "some-repo"} }
+```
+
+- `:severity` : The severity of the alert represents how consumers should support it.
+    - `:error` Should be used if an only if the alert is serious to prevent the
```
rlinehan: s/if an only/if and only/

also seems like "enough" is missing after "serious"?

documentation/developers.md
```diff
@@ -65,6 +65,42 @@ might be implemented to utilize the `compare-levels` function:
...
+      service from being in the `:running` state.
+    - `:warning` Should be used for conditions which will likely result in
+      degraded service without effecting the `:state`
```
rlinehan: s/effecting/affecting/

documentation/developers.md
```diff
@@ -65,6 +65,42 @@ might be implemented to utilize the `compare-levels` function:
...
+    - `:info` : Should be used for other alerts that do not impact the
+      functioning of the service. These should not be present when the status
+      level is greater than info.
+- `:message` : A human readable string that can be displayed to users.
+- `:type` : This can be used by consumer logic to find particular alerts that
+  are important to them.
+- `:details` : Further computer readable information about the cause of the
+  alert. The schema of details should be the same for all alerts of the same
+  type. If multiple alerts of a type are possible they should be distiguishable by details.
```
rlinehan: Seems like it would be good to mention that `type` and `details` are optional?

Also s/distiguishable/distinguishable/

rlinehan:
I left a bunch of nitpicky comments about spelling that don't much matter.

I think it would be good to mention in the docs that `details` and `type` are optional.

I haven't used alerts at all yet, so while this seems fine to me, it seems like it would be a good idea for someone else (@lindboe maybe?) to review this who is more familiar with alerts.

documentation/developers.md
adreyer: I debated that. I would prefer if they were required but that would be a breaking change. I chose to document them as required as a compromise but I guess that could confuse users trying to consume alerts.

nicklewis:
This follows the basic pattern we use for exceptions and HTTP error responses. We use that pattern for a reason. This makes sense to me.

documentation/developers.md
lindboe: What's the intended relation between type/details and information in the status map?

documentation/developers.md
```diff
@@ -65,6 +65,42 @@ might be implemented to utilize the `compare-levels` function:
...
+ :type :my-service/databas-unavailable
+ :details {}}
+
+{:severity :warning
+ :message "Repo may not be initialized"
+ :type :my-service/uninitilized-repo
+ :details { :repo "some-repo"} }
+```
+
+- `:severity` : The severity of the alert represents how consumers should support it.
```
lindboe, line 89: I do honestly prefer severity that is decoupled from availability. There are some potentially severe issues that have nothing to do with whether the service will be able to respond - for example, we might be able to detect a security issue. In the case of replicated services, having broken replication is extremely important to surface (relative importance depends on the app) but will not affect the local service's ability to function.

adreyer: "function" and "respond" are not the same thing. A system that's not replicating will respond but should be marked as unhealthy for alerts and load balancing purposes.

Warnings are still exposed to users and make sense for issues that aren't work marking a box as unhealthy.

cprice404:
Is this blocking any upcoming releases?  What do we need to do to finish it up?  Should we schedule a meeting with a few of the commenters/maintainers + UX to run down the answers to the open questions?

rlinehan:
@adreyer what's the status on this?

rlinehan:
@adreyer @lindboe @nicklewis what's the status of this? anyone?

KevinCorcoran:
bueller

rlinehan:
I'm gonna close this cuz no one seems to care anymore. Feel free to reopen if you're going to do something about it.

PR #70, opened 2016-12-14
(TK-401) Log service name when error or timeout with service status callback

justinstoller:
This updates the signature of `guarded-status-fn-call` to accept the name of the service being checked. It then uses that name to identify the service when logging issues related to it (timeout, malformed response, other exception). Existing tests seemed to adequately test this change.


cprice404 (approved):

rlinehan (approved):
💯

PR #72, opened 2016-12-15
(PE-13539) Log service version numbers at startup

cprice404:
This commit will cause the status service, during startup, to log
the version numbers for each service that registers a callback.


jpinsonault:
👍

PR #73, opened 2016-12-16
(MAINT) prep for 0.7.0 release

cprice404:
This PR updates the CHANGELOG in preparation for the 0.7.0 release.

It also includes a commit that introduces `clj-parent` and gets rid of some unnecessary transitive dependency resolution info.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -19,8 +19,15 @@
                  [slingshot]
                  [prismatic/schema]
                  [trptcolin/versioneer]
-                 [ring/ring-defaults]
+                 ;; ring-defaults brings in a bad, old version of the servlet-api, which
+                 ;; now has a new artifact name (javax.servlet/javax.servlet-api).  If we
+                 ;; don't exclude the old one here, they'll both be brought in, and consumers
+                 ;; will be subject to the whims of which one shows up on the classpath first.
+                 ;; thus, we need to use exclusions here, even though we'd normally resolve
+                 ;; this type of thing by just specifying a fixed dependency version.
+                 [ring/ring-defaults :exclusions [javax.servlet/servlet-api]]
                  [org.clojure/java.jmx]
+                 [org.clojure/tools.logging]
 
                  [puppetlabs/kitchensink]
                  [puppetlabs/trapperkeeper]
--- a/project.clj
+++ b/project.clj
@@ -2,7 +2,7 @@
 (def ks-version "1.3.0")
 
 
-(defproject puppetlabs/trapperkeeper-status "0.6.1-SNAPSHOT"
+(defproject puppetlabs/trapperkeeper-status "0.7.0-SNAPSHOT"
   :description "A trapperkeeper service for getting the status of other trapperkeeper services."
   :url "https://github.com/puppetlabs/trapperkeeper-status"
   :license {:name "Apache License, Version 2.0"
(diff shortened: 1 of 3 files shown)


project.clj
```diff
@@ -1,70 +1,43 @@
...
-                                  ;; End transitive dependency resolution
-
-                                  [puppetlabs/http-client "0.5.0"]
-                                  [puppetlabs/trapperkeeper ~tk-version :classifier "test"]
-                                  [puppetlabs/trapperkeeper-webserver-jetty9 "1.5.6"]
-                                  [puppetlabs/kitchensink ~ks-version :classifier "test"]]}}
+  :profiles {:dev {:dependencies [[puppetlabs/http-client]
+                                  [puppetlabs/trapperkeeper nil :classifier "test"]
+                                  [puppetlabs/trapperkeeper-webserver-jetty9]
+                                  [puppetlabs/kitchensink nil :classifier "test"]]}}
```
KevinCorcoran: are these `nil` values intentional/correct?

KevinCorcoran (approved):
just one question

project.clj
cprice404: They were intentional, they were required for some versions of lein, not sure if the fix for that is in lein 2.7.1 or not.  I'll try it w/o them real quick.

cprice404: yep, works without them, added a commit that removes them.

project.clj
```diff
@@ -1,70 +1,43 @@
...
+  :parent-project {:coords [puppetlabs/clj-parent "0.2.5"]
+                   :inherit [:managed-dependencies]}
 
-                 ;; Dependencies which resolve version conflicts via
-                 ;; :pedantic? :abort in transitive dependencies
-                 [clj-time "0.11.0"]
-                 [ring "1.4.0"]
-                 [commons-codec "1.10"]
-                 [org.clojure/tools.macro "0.1.5"]
-                 [org.clojure/tools.logging "0.3.1"]
```
camlow325, line 23: Should we still have a managed dep listed for this since we have some code which has a `require` for it?

project.clj
```diff
@@ -1,70 +1,43 @@
...
-                 [puppetlabs/trapperkeeper ~tk-version]
-                 [puppetlabs/trapperkeeper-scheduler "0.0.1"]
-                 [puppetlabs/ring-middleware "1.0.0"]
-                 [puppetlabs/comidi "0.3.1"]
-                 [puppetlabs/i18n "0.4.3"]]
+                 [cheshire]
+                 [slingshot]
+                 [prismatic/schema]
+                 [trptcolin/versioneer]
+                 [ring/ring-defaults]
```
camlow325: Do you think we should list the `:exclusions` for `servlet-api` here or not.  I did in my [trapperkeeper-metrics PR](https://github.com/puppetlabs/trapperkeeper-metrics/pull/38) out of paranoia but it may not be necessary.

camlow325 (commented):
Couple of small questions about version deps but I'm good with this if you don't see a need to make any other changes.

project.clj
cprice404: ah, yeah

project.clj
cprice404: sure

cprice404:
Added a commit that does those two things

camlow325 (approved):

PR #74, opened 2017-01-14
(MAINT) Avoid a schema error during an early shutdown

camlow325:
Previously, if a startup error were encountered before the status
service's init function was called, a call to the stop function could
throw/log a schema error when trying to manipulate state that had not
been set on the context during the init call.

With this commit, the status service just avoids the code which would
reset context state that doesn't exist, avoiding the appearance of
the schema error.


camlow325:
I didn't bother adding a unit test for this one since it seemed so trivial.  Happy to do one, though, if others see value in it.

PR #75, opened 2017-01-17
(maint) use schema/Num to validate results of division by available p…

wkalt:
…rocessors

Depending on the number of processors the result may not be an integer.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/status/cpu_monitor.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/cpu_monitor.clj
@@ -11,8 +11,8 @@
 
 (def CpuUsageSnapshot
   {:snapshot {:uptime schema/Int
-              :process-cpu-time schema/Int
-              :process-gc-time schema/Int}
+              :process-cpu-time schema/Num
+              :process-gc-time schema/Num}
    :cpu-usage schema/Num
    :gc-cpu-usage schema/Num})
 


camlow325 (approved):

haus:
@wkalt i presume this means we have no tests that actually use this map?

camlow325:
@haus We have some tests that use the map, like [here](https://github.com/puppetlabs/trapperkeeper-status/blob/0.7.0/test/puppetlabs/trapperkeeper/services/status/status_service_test.clj#L515-L566).  I didn't see any production code which uses it, though.  I suppose we could somehow mock the num-cpus / activity which is calculated so that we can trigger a non-whole number being set for `:process-cpu-time` / `:process-gc-time`.  Not sure if that's worthwhile, though, given that it appears that this is only used for testing.  I didn't see anything in the production code which uses the `:snapshot` after the values are stuffed into the context for it.  The status service endpoint doesn't return them in its payload.

haus:
@camlow325 Cool. Seems fine as is.

jpinsonault:
This seems good to me. I'm not super familiar with this, I'm assuming it doesn't matter if the value becomes a float after division. Looks like the process cpu time bean comes out as a `long` to begin with. Would it be better to keep it as an integer type rather than make the schema more permissive? Honest question, I don't know

camlow325:
> Looks like the process cpu time bean comes out as a long to begin with. Would it be better to keep it as an integer type rather than make the schema more permissive?

We could coerce the values after division by num-cpus back into a long to satisfy the existing schema.  Given that we're apparently not actually using the values for anything outside of testing anyway, though, not sure it matters.  Just updating the schema to reflect what the values already can be seems okay to me.

jpinsonault:
Cool 👍 fine by me

PR #82, opened 2017-07-18
(MAINT) Fix bug in status logging docs

jpinsonault:
The status logger needs to be at the "debug" level in order for it to pick up
the logging events created by the status service, since it logs at debug level

Otherwise the status logger wouldn't log unless the root logging level is set
to debug


justinstoller:
In the spirit of no good dead goes unpunished - would it make sense to mention the necessity of the debug log level and/or additivity=false attributes?

jpinsonault:
Yeah I can put a blurb in there about them

PR #83, opened 2017-09-01
(maint) Update clj-parent to 1.4.3

Sharpie:
This patch updates the clj-parent dependency to 1.4.1 and drops a now-obsolete
exclusion of javax.servlet/servlet-api from the ring/ring-defaults
dependency.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: clojure
 lein: 2.7.1
 jdk:
-- oraclejdk7
-- openjdk7
+  - oraclejdk8
+  - openjdk8
 script: ./ext/travisci/test.sh
 notifications:
   email: false
--- a/project.clj
+++ b/project.clj
@@ -6,7 +6,7 @@
 
   :min-lein-version "2.7.1"
 
-  :parent-project {:coords [puppetlabs/clj-parent "0.6.1"]
+  :parent-project {:coords [puppetlabs/clj-parent "1.4.3"]
                    :inherit [:managed-dependencies]}
 
   :pedantic? :abort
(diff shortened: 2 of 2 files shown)


justinstoller:
https://github.com/puppetlabs/trapperkeeper-status/pull/84 should fix the error seen in this PR

Magisus:
That PR is merged, this will need a rebase.

Sharpie:
Hmm. Looks like the test is still failing after rebase, but I can't reproduce locallay using `lein test`. Failure looks related to the status logging added in [TK-400](https://tickets.puppetlabs.com/browse/TK-400)

Sharpie:
@justinstoller @Magisus Looks like tests were failing due to Travis still using Java 7 as the test platform. I added another commit to bump the tests to Java 8.

Magisus (approved):

justinstoller:
Do we want to update this to tk-status 2.0 since we're dropping Java 7 support?

Sharpie:
Looks like we went with a .y bump for the same sort of change in puppetlabs/trapperkeeper-metrics#52. Java 7 support has been functionally dead for a while now given that clj-parent pulls in components that won't work with it --- and that's what the top-level projects have been running with.

justinstoller:
I don't think I agree with what we did there. I think Jeremy [originally meant](https://github.com/puppetlabs/trapperkeeper-metrics/pull/52#issuecomment-307971433) for us to bump the Y version but not drop Java 7 support, only drop testing of it, and if it breaks then add it back in a Z. I think we then later came back around and released it, but wrote that we were dropping Java 7 support without thinking much about it.

I would prefer that we followed the approach taken with [tk-jetty9](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/commit/3128c2b9ebf97dc3ae49ed21574c76caa8b87e50) and [server](https://tickets.puppetlabs.com/browse/SERVER-1765) and bump the X release. Not the approach taken the first week I was back on the team (and when Jeremy and Ruth were on their way out).

PR #85, opened 2017-10-09
(TK-454) Add last GC duration to gc-stats

Sharpie:
This patch adds the duration from the LastGcInfo attribute of
GarbageCollectorMXBean entries to the gc-stats returned by the base status
service. This info is useful for detecting increased GC activity due to a
leak or poorly optimized code.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/status/status_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_core.clj
@@ -84,7 +84,8 @@
 
 (def GcStatsV1
   {schema/Str {:count schema/Int
-               :total-time-ms schema/Int}})
+               :total-time-ms schema/Int
+               (schema/optional-key :last-gc-info) {:duration-ms schema/Int}}})
 
 (def JvmMetricsV1
   {:heap-memory MemoryUsageV1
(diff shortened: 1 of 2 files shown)


test/puppetlabs/trapperkeeper/services/status/status_core_test.clj
```diff
@@ -124,10 +124,7 @@
           (is (every? #(or (< 0 %) (= -1 %)) (vals (:non-heap-memory jvm-metrics))))
           (is (= #{:max :used} (ks/keyset (:file-descriptors jvm-metrics))))
           (is (every? #(< 0 %) (vals (:file-descriptors jvm-metrics))))
-          (is (every? #(= #{:count :total-time-ms} (ks/keyset %))
+          (is (every? #(= #{:count :total-time-ms :last-gc-info} (ks/keyset %))
                       (vals (:gc-stats jvm-metrics))))
-          (is (every? #(<= 0 %)                             ;; Possible that no major collections occurred.
-                      (mapcat #(vals %)
-                              (vals (:gc-stats jvm-metrics)))))
```
Sharpie: I dropped this test as Schema is doing a good job of enforcing assertions about the structure and data type of the response, re-writing the same assertions in a test is a lot of effort just to duplicate the same checks.

test/puppetlabs/trapperkeeper/services/status/status_core_test.clj
```diff
@@ -124,10 +124,6 @@
           (is (every? #(or (< 0 %) (= -1 %)) (vals (:non-heap-memory jvm-metrics))))
           (is (= #{:max :used} (ks/keyset (:file-descriptors jvm-metrics))))
           (is (every? #(< 0 %) (vals (:file-descriptors jvm-metrics))))
-          (is (every? #(= #{:count :total-time-ms} (ks/keyset %))
-                      (vals (:gc-stats jvm-metrics))))
-          (is (every? #(<= 0 %)                             ;; Possible that no major collections occurred.
-                      (mapcat #(vals %)
-                              (vals (:gc-stats jvm-metrics)))))
+          (is (contains? jvm-metrics :gc-stats))
```
Sharpie, line 127: I loosened these tests as Schema is doing a good job of enforcing assertions about the structure and data types of the response. Re-writing the same assertions in tests is a lot of effort just to produce a fragile duplication of checks which already exist.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -84,7 +84,8 @@
 
 (def GcStatsV1
   {schema/Str {:count schema/Int
-               :total-time-ms schema/Int}})
+               :total-time-ms schema/Int
+               (schema/optional-key :last-gc-info) {:duration-ms schema/Int}}})
```
justinstoller, line 88: I assume since we're only adding an optional field (but not removing anything) we can continue to consider this part of the v1 schema

justinstoller (approved):

src/puppetlabs/trapperkeeper/services/status/status_core.clj
Sharpie: It's added as optional as it is possible for last-gc-info to be `nil` if no collection has occured yet. For example, with Puppet Server, the Major GC algorithm usually doesn't have to fire off a major collection sweep until an hour or so of uptime.

Sharpie: Also, `last-gc-info` could be specific to Oracle's HotSpot JVM as it's not part of the GarbageCollectorMXBean interface. That's what PE uses, but I also flagged it as optional just in case someone is running Open Source Puppet Server or PuppetDB on a different JVM.

PR #86, opened 2017-10-13
(TK-455) Backport service name reporting on timeout to 0.6 

justinstoller:
This brings back
  - (TK-401) Log service name when error or timeout with service status callback

I believe our CI pipelines have been updated to auto generate the pot files that will need to be changed with this. However, I will double check that and may possibly need to bring back ddf96a7
 as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: clojure
-lein: lein2
+lein: 2.7.1
 jdk:
 - oraclejdk7
 - openjdk7
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## 0.6.1
+
+This is a minor release. It includes
+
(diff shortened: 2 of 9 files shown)


justinstoller:
Looks like we need to bring back the travis update as well...

Check CI and the Jenkins job exists but will only work on projects that have the i18n plugin at version 0.5 or higher (so not this branch). Will update this with the commits for travis and i18n in a bit.

justinstoller:
Actually, It doesn't make any sense to not double up this and the release work. As part of the backport we'll need to release an 0.6.1 of tk-status and then pull that into the pe-pse and friends. I'll add the release prep onto this PR instead of doing multiple PRs.

justinstoller:
ping @puppetlabs/puppetserver-maintainers for review

MosesMendoza (approved):

MosesMendoza:
matches up pretty well with https://github.com/puppetlabs/trapperkeeper-status/pull/70/files

PR #89, opened 2018-06-28
(PE-24670) Make status service respect tk-auth configs

amysahli:
Wrap status service routes so that access to the
status/v1/services endpoint can be configured to require
cert-authentication. If no authorization service is running,
the status service will default to allow unauthenticated
requests.

### DIFF
--- a/documentation/developers.md
+++ b/documentation/developers.md
@@ -92,6 +92,26 @@ web-router-service: {
 For information on proxying plaintext `/status` requests to an otherwise HTTPS
 protected server, see the [Status Proxy documentation](./status-proxy-service.md).
 
+Applications using trapperkeeper can require authentication for the status endpoint.
+For this to happen, tk-auth must be included in the application's bootstrap config
+file.  Then, to enable authentication, the application's tk-auth
+configuration file needs to be updated to include this authorization rule:
+
+```
+{
+   # Disallow unauthenticated access to the status service endpoint
+    match-request: {
+        path: "/status/v1/services"
+        type: path
+        method: get
+    }
+    allow-unauthenticated: false
+    allow: "*"
+    sort-order: 500
+    name: "puppetlabs status service"
+}
+```
+
 ## Details
 
 See [Query API](./query-api.md) and [Wire Format](./wire-formats.md) for details
--- a/test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
+++ b/test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
@@ -10,15 +10,30 @@
             [puppetlabs.trapperkeeper.services.status.status-service :refer [status-service get-status]]
             [puppetlabs.trapperkeeper.services.status.status-core :as status-core]
             [puppetlabs.trapperkeeper.services.webrouting.webrouting-service :as webrouting-service]
+            [puppetlabs.trapperkeeper.services.authorization.authorization-service :as tk-auth]
             [puppetlabs.trapperkeeper.services.webserver.jetty9-service :as jetty9-service]
             [puppetlabs.trapperkeeper.services.scheduler.scheduler-service :as scheduler-service]
             [puppetlabs.kitchensink.core :as ks]))
 
 (use-fixtures :once schema-test/validate-schemas)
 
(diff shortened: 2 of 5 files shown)


justinstoller:
We should do a quick update to the docs on how consumers of the status service can use/need to be aware of its interactions with tk-auth. Maybe an example auth.conf in [exposing-the-status-endpoint](https://github.com/puppetlabs/trapperkeeper-status/blob/master/documentation/developers.md#exposing-the-status-endpoint)?

documentation/developers.md
```diff
@@ -78,6 +78,7 @@ webserver: {
     ssl-key: /etc/ssl/private_keys/myhostname.pem
     ssl-ca-cert: /etc/ssl/certs/ca.pem
     default-server: true
+    client-auth: "want"
```
amysahli: @justinstoller is this needed?

documentation/developers.md
```diff
@@ -88,10 +89,29 @@ web-router-service: {
   }
 }
 ```
-
 For information on proxying plaintext `/status` requests to an otherwise HTTPS
 protected server, see the [Status Proxy documentation](./status-proxy-service.md).
 
+The status service can interact with the tk-auth service if it is also a
+dependency. If it is a dependency, it is possible to require authentication for
+the status endpoint.  You can do this by updating the puppetserver/auth.conf file
```
amysahli: ^^ is puppetserver/auth.conf correct?

documentation/developers.md
```diff
@@ -88,10 +89,29 @@ web-router-service: {
   }
 }
 ```
-
 For information on proxying plaintext `/status` requests to an otherwise HTTPS
 protected server, see the [Status Proxy documentation](./status-proxy-service.md).
 
+The status service can interact with the tk-auth service if it is also a
+dependency. If it is a dependency, it is possible to require authentication for
```
justinstoller: It'd be good to distinguish between jar/library dependencies (which tk-auth is now one of for tk-status) vs services loaded during application runtime. Maybe something like, "if it's included in the trapperkeeper application's bootstrap config" rather than "dependency".

documentation/developers.md
```diff
@@ -88,10 +89,29 @@ web-router-service: {
...
+the status endpoint.  You can do this by updating the
+/etc/puppetlabs/puppetserver/conf.d/auth.conf file to include this authorization rule:
```
justinstoller: I'd just say something like "tk-auth's configuration file" or something less specific than the actual path to where Puppet Server puts its tk-auth file (assuming the primary consumers of this documentation are either people outside the Puppet Server team within Puppet or community members using tk-status as a library in their own applications).

documentation/developers.md
justinstoller: This was required for your test that created an https request but didn't provide it's own cert. (we wanted to verify it got a 403 rather than https error). I don't believe it's required for this documentation.

test/puppetlabs/trapperkeeper/services/status/status_service_test.clj
```diff
@@ -192,6 +207,61 @@
...
+                                   {:path "/status/v1/services"
+                                    :type "path"
+                                    :method "get"}
+                                   :sort-order 500
+                                   :allow "*"
+                                   :allow-unauthenticated false
+                                   :name "puppetlabs status service"}]}})
+        (testing "without authentication returns 403"
+          (let [resp (http-client/get "https://localhost:8180/status/v1/services" incomplete-ssl-config)]
+            (is (= 403 (:status resp)))))
```
justinstoller: This causes an error message logged during the test run. Can we wrap this with a `with-test-logging` block to keep that from happening?

documentation/developers.md
amysahli: It's not included in the trapperkeeper application's bootstrap config, is it?  I thought it was in the application's (like puppetserver or puppetdb) bootstrap config file.

justinstoller: I don't think that's a distinction I meant to make, I used "trapperkeeper application" as a generalization of puppetdb, puppetserver, etc. ...Applications using trapperkeeper.

justinstoller (approved):

Magisus (approved):

PR #94, opened 2022-08-01
(maint) Update clj-parent to 5.1.1; fips config; dynapath; fix core reflection

rbrw:
I started this because I'd noticed what looked like notable reflection cost in some pdb flame graphs while working on escalations, and I thought it'd be easy to fix.  Several yaks later, this pr.

It also turned out that I'd misread the graph, and that further down, the real culprit was the i18n reflection problem that we've fixed in newer versions, so we don't have to keep the status-core reflection fixes, but I thought they might be worthwhile since they're fairly straightforward and not too noisy.

See the individual commit messages for specific details.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -16,6 +16,7 @@
   :dependencies [[org.clojure/clojure]
 
                  [cheshire]
+                 [org.tcrawley/dynapath "1.0.0"]
                  [slingshot]
                  [prismatic/schema]
                  [trptcolin/versioneer]
--- a/project.clj
+++ b/project.clj
@@ -6,7 +6,7 @@
 
   :min-lein-version "2.7.1"
 
-  :parent-project {:coords [puppetlabs/clj-parent "1.7.38"]
+  :parent-project {:coords [puppetlabs/clj-parent "5.1.1"]
                    :inherit [:managed-dependencies]}
 
   :pedantic? :abort
--- a/project.clj
+++ b/project.clj
@@ -36,7 +36,8 @@
                                      :password :env/clojars_jenkins_password
                                      :sign-releases false}]]
 
-  :profiles {:dev {:dependencies [[puppetlabs/http-client]
+  :profiles {:dev {:dependencies [[org.bouncycastle/bcpkix-jdk15on]
+                                  [puppetlabs/http-client]
                                   [puppetlabs/trapperkeeper :classifier "test"]
                                   [puppetlabs/trapperkeeper-webserver-jetty9]
                                   [puppetlabs/kitchensink :classifier "test"]]}}
--- a/src/puppetlabs/trapperkeeper/services/status/status_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/status/status_core.clj
@@ -19,6 +19,9 @@
            (javax.management ObjectName)
            (clojure.lang IFn)))
 
+
+(set! *warn-on-reflection* true)
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Schemas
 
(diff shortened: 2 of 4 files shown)


src/puppetlabs/trapperkeeper/services/status/status_core.clj
```diff
@@ -13,12 +13,15 @@
...
             [puppetlabs.i18n.core :as i18n])
-  (:import (java.net URL)
+  (:import (java.net URI)
            (java.util.concurrent CancellationException)
            (java.lang.management ManagementFactory)
            (javax.management ObjectName)
            (clojure.lang IFn)))
 
+
+(set! *warn-on-reflection* true)
```
jonathannewman, line 23: Do we want this turned on in the shipping code?

project.clj
```diff
@@ -16,6 +16,7 @@
   :dependencies [[org.clojure/clojure]
 
                  [cheshire]
+                 [org.tcrawley/dynapath "1.0.0"]
```
mwaggett, line 19: Looks like we have this in clj-parent: https://github.com/puppetlabs/clj-parent/blob/7d37cd237cbbe8e93b461483b1d95c286829800c/project.clj#L91

Should we just bump the version there?

mwaggett (commented):
The changes mostly look good to me!
Some of the commit messages could be clearer though, in my opinion:
* The `thing: actual change` format is a bit weird to me. I find it clearer to say something like `Made change in thing`. Does that make sense? 
* The multi-line error commit could use a little more info - the behavior of what changed? when?
* I would find it useful for the reflection commit to explain why the change is still worthwhile, but that could just be me not knowing anything about reflection.

project.clj
rbrw: I'm not sure, I think maybe Austin told me that there was some existing issue with puppetserver that would break if we did, and the fix hasn't been prioritized yet?  It's currently also pinned in pdb, to fix the same issue, fwiw.

rbrw:
> The changes mostly look good to me! Some of the commit messages could be clearer though, in my opinion:
> 
>     * The `thing: actual change` format is a bit weird to me. I find it clearer to say something like `Made change in thing`.

Ahh, don't feel strongly about it, but I'm used to using either in various projects, and I sometimes favor the former to either preserve space on the summary line, or to keep a series of related changes to a given subsystem easier to spot in a `git log --oneline --decorate --graph` or similar.

>     * The multi-line error commit could use a little more info - the behavior of what changed? when?

Hmm, which one?  You mean the `.toArray` fix?  If so, that was just a bug in `ordered` with respect to, and if I recall correctly, newer clojure and/or jdk versions (not sure of the precise details offhand, but I can do some research if we'd like -- just recalled addressing it in pdb a while back, so I knew what to do).

>     * I would find it useful for the reflection commit to explain why the change is still worthwhile, but that could just be me not knowing anything about reflection.

It's mostly just that reflection can be very expensive if it's ever involved in fine-grain enough use (the canonical example might be a tight loop), and so it's never desirable, but the annotations add "noise"/maint, and so might not be worth it if there are a lot of them in code that's always coarse grain.  Here the changes were trivial enough that I was leaning toward "won't hurt, might help", since I'd already done them, but I can just drop that commit if we prefer.

src/puppetlabs/trapperkeeper/services/status/status_core.clj
rbrw: No strong opinion - I'd been thinking that either we'd keep the file clean now that it's clean, or, we'd decide we didn't want to at some point later, and we'd remove the line then.  I don't think it should have any affect in production otherwise, offhand?

rbrw: (Oh, and to be clear, also happy to just take that out.)

mwaggett:
@rbrw Is this something you're still interested in landing? If so, it could use some updates to the clj-parent version and the bouncycastle jdk15on -> jdk18on change.

PR #96, opened 2023-07-05
(PE-36390) replace use of ring defaults, update deps

steveax:


project.clj
```diff
@@ -36,7 +36,8 @@
                                      :password :env/clojars_jenkins_password
                                      :sign-releases false}]]
 
-  :profiles {:dev {:dependencies [[puppetlabs/http-client]
+  :profiles {:dev {:dependencies [[org.bouncycastle/bcpkix-jdk15on]
```
jonathannewman:
```suggestion
  :profiles {:dev {:dependencies [[org.bouncycastle/bcpkix-jdk18on]
```
We are trying to eliminate the old `15on` in favor of the `18on` which is the newer version.

jonathannewman (approved):
Small nit about bouncy castle dev version, but LGTM!

